PR #41, opened 2022-03-09
Add Find() method to schema registry

flimzy:
Relates to https://github.com/invopop/gobl.cli/pull/7

### DIFF
--- a/schema/register.go
+++ b/schema/register.go
@@ -1,6 +1,9 @@
 package schema
 
-import "reflect"
+import (
+	"path/filepath"
+	"reflect"
+)
 
 // registry contains all the schemas that we can possibly know about from either
 // inside or outside GOBL.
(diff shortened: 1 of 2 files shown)


flimzy:
@samlown What do you think about this type of approach?

The idea with the confidence return variable is that we could classify a fuzzy match in some way, and if the value falls too low (< 0.90?) we can return a "no match" type of error.

In the future, we might add another method that returns a list of matches, ordered by confidence, for display in a UI of some sort.

Still to do:

- [ ] Actually do the case folding match against the schema name
- [ ] Select the newest version when there are multiple
- [ ] Add a fuzzy text match library??

samlown:
Okay, interesting! From a user perspective, it could be really useful to just put in `schema.Find("invoice")` and get a result back.

I'm not totally sure if this should be in this lib or just in the cli. There is an argument to be made that `gobl` should be as precise as possible to avoid potentially ambiguous results, especially if you consider future additions.

To give a lame example, but imagine `schema.Find("bill")` gives back an Invoice. If we add or expose some other object to bill in the future, that might suddenly stop working or provide something else.

flimzy:
This functionality hsa been moved into https://github.com/invopop/gobl.cli/pull/7

PR #43, opened 2022-03-24
Run tests against WASM

flimzy:


samlown (approved):
That looks very straight forward!

PR #44, opened 2022-03-30
POC for validator that works with GopherJS

flimzy:
This fixes one of the failing validation tests in GopherJS.  If we like this approach, I'll finish replacing `github.com/go-ozzo/ozzo-validation` with `github.com/go-playground/validator/v10`, and then we may be able to run this entire thing with GopherJS.


samlown (commented):
A problem I see with this approach is the lack of strict typing in the validation definitions, and the fact they could get pretty long.

Ozzo also has a clever mechanism that will call `Validate()` on all sub-models, which the go-playground version doesn't appear to support.

I'd be in favour of considering defining our own simplified validator package tailor made for GOBL, based on the Ozzo model of implementing the `Validatable` interface and thus permitting a gradual conversion. I'd call the package `validate` and define wrapper aournd `is` so it reads better and has better naming, e.g.:

```
import (
  "github.com/invopop/gobl/validate"
  "github.com/invopop/gobl/validate/is"
)

func (r Region) Validate() error {
  return validate.Struct(&r,
    validate.Field(&r.Code, is.Required),
  )
}
```

samlown:
Also, the Ozzo library doesn't appear to be all that well maintained: https://github.com/go-ozzo/ozzo-validation another reason why I like something internal to gobl that covers the basics.

PR #46, opened 2022-05-04
Massive refactor to be able to better support taxes and schemes

samlown:
To many changes here to be reasonable reviewed... Everything is a result of trying to improve support for being able to generate invoices that can be used for exporting goods using tax rates at the destination and adding support for meta data to be included for special providers.

Also included is a switch to a standard approach of constant naming with the "namespace" defined first, followed by the type.

Taxes in line items are now defined as a hash as opposed to an array. The concept of Code and Key has been refined and implemented.


bill/discounts.go
```diff
@@ -67,7 +67,7 @@ func (m *Discount) Validate() error {
 }
 
 // GetTaxRates responds with the array of tax rates applied to this line.
-func (m *Discount) GetTaxRates() tax.Rates {
+func (m *Discount) GetTaxes() tax.Map {
```
flimzy, line 70: Stylistic: Go convention is to omit the word 'Get' from getters.

Also: Why do we need getters for exported fields?  Maybe to satisfy an interface?

bill/invoice.go
```diff
@@ -68,13 +62,13 @@ type Invoice struct {
 	// Expenses paid for by the supplier but invoiced directly to the customer.
 	Outlays Outlays `json:"outlays,omitempty" jsonschema:"title=Outlays"`
 
-	// Summary of all the invoice totals, including taxes.
-	Totals *Totals `json:"totals" jsonschema:"title=Totals"`
-
 	Ordering *Ordering `json:"ordering,omitempty" jsonschema:"title=Ordering Details"`
 	Payment  *Payment  `json:"payment,omitempty" jsonschema:"title=Payment Details"`
 	Delivery *Delivery `json:"delivery,omitempty" jsonschema:"title=Delivery Details"`
```
flimzy, line 67: Point of curiosity: I guess this means there's no way to specify multiple ship-to addresses on a single invoice?

bill/invoice.go
```diff
@@ -68,13 +62,13 @@ type Invoice struct {
 	// Expenses paid for by the supplier but invoiced directly to the customer.
 	Outlays Outlays `json:"outlays,omitempty" jsonschema:"title=Outlays"`
 
-	// Summary of all the invoice totals, including taxes.
-	Totals *Totals `json:"totals" jsonschema:"title=Totals"`
-
 	Ordering *Ordering `json:"ordering,omitempty" jsonschema:"title=Ordering Details"`
 	Payment  *Payment  `json:"payment,omitempty" jsonschema:"title=Payment Details"`
```
flimzy, line 66: Curiosity: Also, no possibility for multiple payment methods?  (part in cash, part credit, for example)

bill/invoice.go
samlown: No, it's not possible to define multiple delivery addresses. I suspect however that this isn't a very widely demanded feature. Normally you'd have purchase orders and delivery notes with details about multiple locations, the invoice would just be the end result of those.

The EU EN 16931-1 standard electronic invoicing definition also has this limitation.

bill/invoice.go
samlown: Kind of... it's possible to define a list of acceptable methods, but only possible to define a single preferred method.

You have `pay.Instructions` for that: https://github.com/invopop/gobl/blob/main/pay/instructions.go#L42

PR #49, opened 2022-05-19
Improved tax errors

samlown:
* Quick set of changes to provide more useful tax usage error messages.

### DIFF
--- a/tax/errors.go
+++ b/tax/errors.go
@@ -1,0 +1,30 @@
+package tax
+
+import "fmt"
+
+// Error is a general wrapper around tax errors produced during run
+// time, typically during calculations. Not to be confused with errors
+// produced from definition validation.
+type Error struct {
+	Reason Key
+}
+
+// Standard list of tax errors
+var (
+	ErrInvalidCategory      = &Error{Reason: "invalid-category"}
+	ErrInvalidRate          = &Error{Reason: "invalid-rate"}
+	ErrInvalidDate          = &Error{Reason: "invalid-date"}
+	ErrInvalidPricesInclude = &Error{Reason: "invalid-prices-include"}
+)
+
+// Error serializes the error's message.
+func (e Error) Error() string {
+	return e.Reason.String()
+}
+
+// WithMessage wraps around the original error so we can use if for matching
+// and adds a human message.
+func (e *Error) WithMessage(msg string, s ...interface{}) error {
+	msg = fmt.Sprintf(msg, s...)
+	return fmt.Errorf("%w: %v", e, msg)
+}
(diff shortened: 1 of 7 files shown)


tax/region.go
```diff
@@ -182,15 +181,15 @@ func (r *Rate) On(date cal.Date) *RateValue {
 func (r *Region) prepareCombo(c *Combo, date cal.Date) error {
 	c.category = r.Category(c.Category)
 	if c.category == nil {
-		return fmt.Errorf("failed to find category, invalid code: %v", c.Category)
+		return ErrInvalidCategory.WithMessage("'%s'", c.Category.String())
 	}
 	c.rate = c.category.Rate(c.Rate)
 	if c.rate == nil {
-		return fmt.Errorf("failed to find rate definition, invalid code: %v", c.Rate)
+		return ErrInvalidRate.WithMessage("'%s' in category '%s'", c.Rate.String(), c.Category.String())
```
agorer, line 188: Do you mean _not in_?

agorer (approved):

tax/errors.go
```diff
@@ -0,0 +1,30 @@
...
+type Error struct {
+	Reason Key
+}
+
+// Standard list of tax errors
+var (
+	ErrInvalidCategory      = &Error{Reason: "invalid-category"}
+	ErrInvalidRate          = &Error{Reason: "invalid-rate"}
+	ErrInvalidDate          = &Error{Reason: "invalid-date"}
+	ErrInvalidPricesInclude = &Error{Reason: "invalid-prices-include"}
```
flimzy, line 17: Consider using constants, so that these package variables can't be changed at runtime.

Or unexport the `Error` type, then they can't be modified by any code outside of this package.

tax/errors.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// Error serializes the error's message.
+func (e Error) Error() string {
+	return e.Reason.String()
+}
+
+// WithMessage wraps around the original error so we can use if for matching
+// and adds a human message.
+func (e *Error) WithMessage(msg string, s ...interface{}) error {
```
flimzy, line 27: Is there any reason we can't just let the callers call `fmt.Errorf` themselves? That's a well-established convention. A special method begs the question what's different...

tax/region.go
samlown: that would be clearer...

tax/errors.go
samlown: I ended up refactoring this hence why it looks like this. Yes, that would make sense.

tax/errors.go
samlown: I had a different approach before as the message was set on the struct. `WithMessage` means we could go back to that approach if needed and we get consistent formatting...

flimzy: YAGNI? :)

flimzy (approved):

tax/errors.go
samlown: haha true, but I still like the formatting :-)

PR #54, opened 2022-06-10
Refactoring surcharge handling, removing VATEQS from Spain

samlown:
* Eliminate the Spanish Category "VAT Equivalence Surcharge"
* Include `surcharge` option inside tax definitions, combos, and totals.
* Ensure surcharges are added when needed to calculations.


regions/fr/fr.go
```diff
@@ -0,0 +1,70 @@
+package gb
```
agorer, line 1: ???

agorer (approved):

PR #55, opened 2022-06-13
Remove included tax rate from all prices

samlown:
* Provides a method on the `Invoice` that will automatically convert all invoice lines, charges, and discounts that include the "price includes tax" category, to versions that do not.
* Very useful for services that do not support prices including tax.
* Supports an `accuracy` property in the `RemoveIncludedTaxes` method that will allow the new precision to be set.
* Doesn't modify the invoice in-place, instead provides a new complete invoice.


samples/es/invoice-es-es-freelance.json
```diff
@@ -1,7 +1,7 @@
 {
 	"$schema": "https://gobl.org/draft-0/envelope",
 	"head": {
-		"uuid": "8cb8f9e3-e8c7-11ec-96bb-e2c0624703d9",
+		"uuid": "c68abf80-eb63-11ec-9fd9-e2c0624703d9",
```
agorer, line 4: why do we need all this changes of UUID and signatures?

agorer (approved):

bill/discounts.go
```diff
@@ -76,3 +76,18 @@ func (m *Discount) GetTaxes() tax.Set {
 func (m *Discount) GetTotal() num.Amount {
 	return m.Amount.Invert()
 }
+
+func (m *Discount) removeIncludedTaxes(cat tax.Code, accuracy uint32) *Discount {
```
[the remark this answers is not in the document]
samlown: Yes, but Go doesn't make that easy to do (generics). I've refactored the amount handling with new `Upscale` and `Remove` methods to at least reduce the amount of code.

samples/es/invoice-es-es-freelance.json
samlown: They're updated because of the tests... originally these JSON were ignored. Not sure what to do about it yet as it's useful to have examples in the code.

samlown: okay, figured out a way to do that now and avoid updates.

PR #57, opened 2022-06-20
Extracting Schema extractor, adding general parser

samlown:
* Schema ID extraction from any JSON document is now more generic.
* Added a new `gobl.Parse` method that extracts **any** GOBL structure defined with a known schema. This is useful for when you don't know if you'll need to handle an Envelope or a document payload.
* Root errors now use `key` instead of `code`, which is more consistent with usage in rest of library.

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -1,0 +1,48 @@
+package gobl_test
+
+import (
+	"testing"
+
+	"github.com/invopop/gobl"
+	"github.com/invopop/gobl/note"
+	"github.com/stretchr/testify/assert"
+)
+
+var parseExampleDoc = `{
+	"$schema": "https://gobl.org/draft-0/note/message",
+	"title": "Test Message",
+	"content": "We hope you like this test message!"
+}`
+var parseExampleEnvelope = `{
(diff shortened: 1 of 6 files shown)


parse_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"testing"
+
+	"github.com/invopop/gobl"
+	"github.com/invopop/gobl/note"
+	"github.com/stretchr/testify/assert"
+)
+
+var parseExampleDoc = `{
+	"$schema": "https://gobl.org/draft-0/note/message",
```
agorer, line 12: Can we upload this schemas to the _gobl.org_ website? It will help code editors when opening a GOBL file.

agorer (approved):

parse_test.go
samlown: I haven't been able to figure out how to do it yet! The gobl.org site is auto-generated, but ideally we want HTML or JSON responses depending on the `Accept` header in the request... I'll see if we can at least provide JSON responses for the time being, ignoring the file extension of the source.

samlown: Bit of a hack, but this is now live!

PR #58, opened 2022-07-01
Fix email validation, only check syntax and not DNS

dstotijn:
Email addresses of a party are [validated](https://pkg.go.dev/github.com/invopop/gobl/org#Email.Validate). We [use](https://github.com/invopop/gobl/blob/v0.26.1/org/party.go#L131) the [`is.Email`](https://pkg.go.dev/github.com/go-ozzo/ozzo-validation/v4/is#Email) validator, which checks syntax but _also_ does this:

> “It also checks if the MX record exists for the email domain.”

With `GOOS=js` and `GOARCH=wasm`, because there is no local DNS resolver when `net.LookupMX` is used, validation fails for valid email addresses.

This PR replaces usages of `is.Email` with [`is.EmailFormat`](https://pkg.go.dev/github.com/go-ozzo/ozzo-validation/v4/is#EmailFormat). That way we still validate the syntax of email addresses, but just don’t validate that the domain has valid MX record(s).

### DIFF
--- a/org/party.go
+++ b/org/party.go
@@ -128,7 +128,7 @@ func (p *Party) Validate() error {
 // Validate ensures email address looks valid.
 func (e *Email) Validate() error {
 	return validation.ValidateStruct(e,
-		validation.Field(&e.Address, validation.Required, is.Email),
+		validation.Field(&e.Address, validation.Required, is.EmailFormat),
 	)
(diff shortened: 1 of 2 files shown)


samlown (approved):
Nice find!

PR #59, opened 2022-07-01
Locality key code refactor

samlown:
* Main objective was to add pre-defined "localities" to region definitions, specifically for Spain whereby some provinces have specific local requirements. This is specifically needed for the TicketBAI integration.
* Ended up refactoring the Code and Key types which had been duplicated a few times into a single `org.Code` and `org.Key` ready to be used throughout the project.
* Modified the `org.Meta` map to require an `org.Key`, thus forcing all meta fields to comply with basic key formats.

### DIFF
--- a/org/code.go
+++ b/org/code.go
@@ -1,26 +1,28 @@
-package tax
+package org
 
 import (
 	"regexp"
 
 	validation "github.com/go-ozzo/ozzo-validation/v4"
+	"github.com/invopop/jsonschema"
 )
 
 // Code represents a string used to uniquely identify the data we're looking
 // at. We use "code" instead of "id", to reenforce the fact that codes should
 // be more easily set and used by humans within definitions than IDs or UUIDs.
-// Tax codes are standardised so that when validated they must contain between
+// Codes are standardised so that when validated they must contain between
 // 2 and 6 inclusive upper-case letters or numbers.
 type Code string
 
 var (
-	codeValidationRegexp = regexp.MustCompile(`^[A-Z0-9]{2,6}$`)
+	codePattern          = `^[A-Z0-9]{1,6}$`
+	codeValidationRegexp = regexp.MustCompile(codePattern)
 )
 
 // Validate ensures that the code complies with the expected rules.
 func (c Code) Validate() error {
 	return validation.Validate(string(c),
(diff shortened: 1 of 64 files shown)


org/code.go
```diff
@@ -1,26 +1,28 @@
...
-	codeValidationRegexp = regexp.MustCompile(`^[A-Z0-9]{2,6}$`)
+	codePattern          = `^[A-Z0-9]{1,6}$`
+	codeValidationRegexp = regexp.MustCompile(codePattern)
 )
 
 // Validate ensures that the code complies with the expected rules.
 func (c Code) Validate() error {
 	return validation.Validate(string(c),
-		validation.Length(2, 6),
+		validation.Length(1, 6),
```
dstotijn, line 25: I think we can omit this rule, because the regex pattern already includes the `{1,6}` quantifier. Alternatively, for consistency's sake, we could add a `{2,64} quantifier to the regexp pattern of `org.Key`.

org/meta.go
```diff
@@ -10,4 +10,4 @@ package org
 //
 // We need to always use strings for values so that meta-data is easy to convert
 // into other formats, such as protobuf which has strict type requirements.
-type Meta map[string]string
+type Meta map[Key]string
```
dstotijn, line 13: Just to double check the semantics of this type: Is metadata of orgs always related to an `org.Key` value?

dstotijn (approved):
Looks good! 👍  Left two minor comments.

org/code.go
samlown: The main objetive was having the extra verbosity of the output, but you're right, it's not good to duplicate. I'll alter the regexp I think.

org/meta.go
samlown: That's the objective, yes. Key has a specific format, so the idea is to make the keys of meta-data more consistent, and a bit easier to define as constants in the code. Here is an example usage: https://github.com/invopop/gobl/blob/f49eb172e84a70fadd316bd7acc2b047253c1321/regions/es/es.go#L136

It could have been done with a String, but this feels easier to check... opinions welcome if you think its overdone!

dstotijn: In that case I think `Key` is perfectly fine 👍

PR #63, opened 2022-07-22
Testing out options for pre-defined keys

samlown:
* Major refactor of keys and codes to ensure that when possible the JSON Schema will be able to make suggestions about the possible values.
* Most usage of the "Key" parameter has been returned to a previous state of having a unique model/type for each Key or Code type.


samlown:
Will merge this, but comments still welcome!

PR #65, opened 2022-08-03
Fix nl tax code

samlown:
* Discovered that the NL tax code validation was not working for all examples (as far as I can gather) from a change in 2020 that meant codes need to be checked using either a mod11 or mod97 algorithm. We only had mod11.


samlown:
I'll merge this directly. But you might find it interesting @dstotijn that we weren't correctly able to verify your NL Tax ID :-)

PR #71, opened 2022-08-30
Experimenting with event defenition

samlown:
* Quick experiment to incorporate Events and Times into the `cal` package.


samlown:
Events are cool, but we don't have a specific use-case yet.

PR #73, opened 2022-09-21
Removing array types for simpler references

samlown:
* Relatively simple change to remove simple array types and instead use regular arrays.
* Fixes an issue whereby the complete notes were duplicated between the `org` and `tax` packages.

### DIFF
--- a/build/schemas/tax/region.json
+++ b/build/schemas/tax/region.json
@@ -71,142 +71,6 @@
       ],
       "description": "Locality represents an area inside a region, like a province or a state, which shares the basic definitions of the region, but may vary in some validation rules."
     },
-    "Note": {
-      "properties": {
-        "key": {
-          "$ref": "#/$defs/NoteKey",
-          "title": "Key",
-          "description": "Key specifying subject of the text"
-        },
-        "code": {
-          "type": "string",
-          "title": "Code",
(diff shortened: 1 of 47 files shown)


build/schemas/tax/region.json
```diff
@@ -344,7 +208,7 @@
           "description": "List of tax category codes that can be used when this scheme is\napplied."
         },
         "note": {
-          "$ref": "#/$defs/Note",
+          "$ref": "https://gobl.org/draft-0/org/note",
           "title": "Note",
           "description": "Notes defines messages that should be added to a document\nwhen this scheme is used."
```
cavalle, line 213: Just noticed that this description was in plural when, at the moment, one single note can be set here. This was pre-existing discrepancy but perhaps we can fix it now?

cavalle (approved):
For what is worth, changes look good to me. I just left one comment about a misleading description that you may want to fix as part of this PR. I've prepared [invopop/](https://github.com/invopop/gobl.ruby/pull/25) in the Ruby library to adapt to these changes.

build/schemas/tax/region.json
samlown: Good find! Yes, I'll fix.

PR #75, opened 2022-10-28
Bump version to release CO region changes

cavalle:
This PR bumps the version of the library to release the new CO region


cavalle:
@samlown could you please confirm this is the right way to release the CO region changes?

cavalle:
Good to know! Next time, I won't regenerate the schemas if there are no changes. Thanks!

PR #76, opened 2022-11-08
Preceding now array, fixed key listings in schema

samlown:
* Bill Invoice `preceding` property now changed to an array in case there are multiple preceding invoices to be modified.
* Upgraded to latest JSON Schema generator
* Correction key options now included in schema.


agorer (approved):

cavalle (approved):
Not an expert in golang or this part of gobl, but the changes look sensible to me!

PR #77, opened 2022-11-23
Use alternative CO tax name

cavalle:
* Apparently, the ReteRenta tax is commonly known as "Retefuente". This is how the DIAN itself calls it in their own PDF invoice representations.

### DIFF
--- a/build/data/tax/CO.json
+++ b/build/data/tax/CO.json
@@ -408,7 +408,7 @@
     {
       "code": "RR",
       "name": {
-        "es": "ReteRenta"
+        "es": "Retefuente"
       },
       "desc": {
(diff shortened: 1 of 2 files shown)


samlown (approved):
Is the code still valid as "RR"? Could also be "RF"?

cavalle:
@samlown as discussed over Slack (replicating here for future reference), we can look at what the DIAN itself does, and they seem to be using "ReteRenta" internally (e.g. in the UBL) and "Retefuente" only for visualisation. So, similarly, while it's not ideal to use different names, I think it's ok to keep using RR internally and leave "Retefuente" as the display name of the tax.

PR #78, opened 2022-11-25
Adding all CO localities at a municipal level

samlown:
* Managed to convert source data using this fab tool: https://tableconvert.com/excel-to-template
* The locality will be made a requirement.


cavalle (approved):
LGTM!

PR #79, opened 2022-11-29
Preceding stamps

samlown:
* Moving `Stamp` from header to `org` package so it can be re-used.
* Adding stamps to bill.Preceding
* Adding new use-case for "revoked" correction method.


cavalle (approved):
LGTM! 🚀

PR #82, opened 2022-12-21
Image support for org party and person

samlown:
Use-Case: be able to include details about a logo so that companies who issue invoices on behalf of others can include images in PDF output, etc.

* First approximation for adding an "Image" structure to the `org` package.
* Added `logos` to `Party`.
* Added `avatars` to `Person`.

The objective of this PR is to add support for being to include company logo details inside the original GOBL. The logo provided for a party can then be used inside PDFs or other output formats instead of the configuration of a provider.

There may be an argument for being able to embed the actual image in Base64, but I'm not sure about that yet. Could work well for SVGs. Thoughts?

### DIFF
--- a/build/schemas/org/image.json
+++ b/build/schemas/org/image.json
@@ -1,0 +1,65 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://gobl.org/draft-0/org/image",
+  "$ref": "#/$defs/Image",
+  "$defs": {
+    "Image": {
+      "properties": {
+        "uuid": {
+          "$ref": "https://gobl.org/draft-0/uuid/uuid",
+          "title": "UUID",
+          "description": "Unique ID of the image"
+        },
+        "label": {
+          "type": "string",
+          "title": "Label",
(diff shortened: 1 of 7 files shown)


cavalle (commented):
Code LGTM and I don't have a strong opinion about whether or not GOBL needs this. I can only contribute a couple of quick thoughts:

* OASIS UBL doesn't seem to support anything like this
* Did you consider using "attachments" for this purpose? I know they aren't part of GOBL but wouldn't it make sense to make them part of GOBL?
* As for the base64 support, I'm with you, it's not necessarily a crazy idea, but my first instinct would be to leave it out until there's a real use case for it.

cavalle (approved):

samlown:
@cavalle you're correct, we're missing context on the problem to solve! Sorry. I'll update the original message, but basically the objective is to be able to support the B2B2B use-case. I.e., a company that has multiple companies, and needs to be able to have a logo on the printed invoice for each.

cavalle:
@samlown thanks! I only caught up with Slack after sending my review here, and then connected the dots about the use case for this PR 😅

PR #83, opened 2022-12-26
Moving to Tax Regimes with country and zone instead of regions and locality

samlown:
* Try to better refine the concepts used for tax identities and addresses.
* Moved "Tax Region" into "Tax Regime"
* A "Tax Locality" inside a given country, is now a "Tax Zone", implying that it may be independent from the locality and region used inside and address.


samlown:
@cavalle @MarkVillacampa this is a bit of an experiment that suddenly occurred to me for refactoring the naming of the different structures related to taxes and addresses. What do you think?

cavalle:
@samlown Just realised the schema needs to be regenerated after the namespace renames to update the JSON files

cavalle:
And the schema version needs to be bumped, if I'm not wrong

samlown:
> And the schema version needs to be bumped, if I'm not wrong

I usually update both the version and the schema after merging with master. I'm still trying to think of the best way to automate this and put the generated schemas somewhere else (main decision point is where to put the output!)

PR #87, opened 2023-01-12
Aligning ordering details with EN16931

samlown:
* After comments about being able to include purchase order details from clients, we realised that GOBL was not completely in sync with the EN16931-1:2017 standard.
* Added `websites` property to `org.Party` which for some reason had been forgotten about!

### DIFF
--- a/bill/payment.go
+++ b/bill/payment.go
@@ -9,8 +9,8 @@ import (
 
 // Payment contains details as to how the invoice should be paid.
 type Payment struct {
-	// The party responsible for paying for the invoice, if not the customer.
-	Payer *org.Party `json:"payer,omitempty" jsonschema:"title=Payer"`
+	// The party responsible for receiving payment of the invoice, if not the supplier.
+	Payee *org.Party `json:"payee,omitempty" jsonschema:"title=Payer"`
 	// Payment terms or conditions.
 	Terms *pay.Terms `json:"terms,omitempty" jsonschema:"title=Terms"`
 	// Any amounts that have been paid in advance and should be deducted from the amount due.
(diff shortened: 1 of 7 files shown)


bill/payment.go
```diff
@@ -9,8 +9,8 @@ import (
 
 // Payment contains details as to how the invoice should be paid.
 type Payment struct {
-	// The party responsible for paying for the invoice, if not the customer.
-	Payer *org.Party `json:"payer,omitempty" jsonschema:"title=Payer"`
+	// The party responsible for receiving payment of the invoice, if not the supplier.
+	Payee *org.Party `json:"payee,omitempty" jsonschema:"title=Payer"`
```
samlown, line 13: Turned out that the "Payee" is the most relevant detail here as opposed to who the payer should be.

org/party.go
```diff
@@ -101,6 +103,18 @@ type Telephone struct {
...
+// Website describes what is expected for a web address.
+type Website struct {
+	// Unique identity code
+	UUID *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
+	// Identifier for this number.
+	Label string `json:"label,omitempty" jsonschema:"title=Label"`
+	// Title of the website to help distinguish between this and other links.
+	Title string `json:"title,omitempty" jsonschema:"title=Title"`
+	// URL for the website.
+	URL string `json:"url" jsonschema:"title=URL,format=uri"`
```
samlown, line 115: Property name could be `URI` instead of `URL`, but I think its okay in this case to more precisely indicate "Location".

org/party.go
```diff
@@ -29,6 +29,8 @@ type Party struct {
 	Addresses []*Address `json:"addresses,omitempty" jsonschema:"title=Postal Addresses"`
 	// Electronic mail addresses
 	Emails []*Email `json:"emails,omitempty" jsonschema:"title=Email Addresses"`
+	//
+	Websites []*Website `json:"websites,omitempty" jsonschema:"title=Websites"`
```
samlown, line 33: The name `Websites` felt the most appropriate here as it feels more explicit compared to something like "links".

cavalle (approved):
Not an expert on the domain or EN 16931-1:2017 but the changes look totally sensible to me!

PR #91, opened 2023-02-15
Heredocs

samlown:
* Using a "heredoc" style approach to add more complete descriptions to the JSON Schema.
* Providing example JSON documents.

### DIFF
--- a/build/schemas/envelope.json
+++ b/build/schemas/envelope.json
@@ -6,7 +6,7 @@
     "Document": {
       "type": "object",
       "title": "Document",
-      "description": "Contents of the envelope that must contain a $schema."
+      "description": "Contains the document payload to be included inside an Envelope.\n\nThe document must contain a \u003ccode\u003e$schema\u003c/code\u003e property that identifies\nthe data's structure otherwise it will be rejected."
     },
     "Envelope": {
       "properties": {
@@ -40,7 +40,7 @@
         "head",
         "doc"
       ],
-      "description": "Envelope wraps around a gobl document and provides support for digest creation and digital signatures."
+      "description": "Envelope wraps around a document adding headers and digital signatures."
     },
     "Header": {
       "properties": {
@@ -91,7 +91,7 @@
         "uuid",
         "dig"
       ],
-      "description": "Header defines the meta data of the body."
+      "description": "Header defines the metadata of the body."
     }
   },
   "$comment": "Generated with GOBL v0.36.0"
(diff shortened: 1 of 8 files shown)


build/schemas/envelope.json
```diff
@@ -6,7 +6,7 @@
     "Document": {
       "type": "object",
       "title": "Document",
-      "description": "Contents of the envelope that must contain a $schema."
+      "description": "Contains the document payload to be included inside an Envelope.\n\nThe document must contain a \u003ccode\u003e$schema\u003c/code\u003e property that identifies\nthe data's structure otherwise it will be rejected."
```
cavalle, line 9: Super-minor thing, but can't those unicode characters be avoided?

build/schemas/note/message.json
```diff
@@ -25,7 +25,13 @@
...
         "content"
       ],
-      "description": "Message represents the minimum possible contents for a GoBL document type."
+      "description": "Message represents a simple message object with a title and some content meant.",
+      "examples": [
+        {
+          "content": "This is an example message.",
+          "title": "Example Title"
+        }
+      ]
```
cavalle, line 34: It's great to have examples! Personally, I like them to look as real as possible. I hate it when I find examples like:` { "first_name": "first name", "last_name": "last name", "vat_code": "99999" }`. Those add little value. A real-looking example provides information about the meaning of the field implicitly, helping readers reinforce their understanding of the semantics.

build/schemas/envelope.json
cavalle, line 9: Minor thing: can't those unicode characters be avoided?

document.go
```diff
@@ -116,8 +117,12 @@ func (d *Document) MarshalJSON() ([]byte, error) {
...
 	return &jsonschema.Schema{
-		Type:        "object",
-		Title:       "Document",
-		Description: "Contents of the envelope that must contain a $schema.",
+		Type:  "object",
+		Title: "Document",
+		Description: here.Doc(`
+			Contains the document payload to be included inside an Envelope.
+			
+			The document must contain a <code>$schema</code> property that identifies
```
cavalle, line 125: I wonder whether we could enter markdown here directly. We don't even need to transform it to html, as it will end up in a MDX file that will take care of that.

cavalle (approved):
ooking good! I left some feedback. Nothing that can't be deferred to future PRs, though.

document.go
samlown: Yeah, let's try with the markdown.

samlown: I think I found a way, a bit messy, but works :-)

PR #92, opened 2023-02-15
Initial support for a party and item 'Identity'

samlown:
* Added a new `Identity` model to the `org` package to be used to define additional IDs that are not associated with tax.
* Defined some common, French, and US, identity type examples.
* Refined the meaning of the `cbc.Code` so that it can be re-used for more identifies and include restrictions to help with normalization.


samlown:
@cavalle comments still welcome on this even after merging!

PR #93, opened 2023-02-17
Add tax code validations for Italy's Partita IVA (P.IVA)

lucanioi:
- Add validations for [P.IVA ](https://it.wikipedia.org/wiki/Partita_IVA)

- The validation assumes that the Tax code is already that of an Italian entity. However, PatturaPA allows the entry of foreign VAT numbers in the same field where P.IVA is expected—when the validation of the invoice happens, how should we support these cases of foreign IDs? For the time being, should we at least check if the ID is Italian and skip the validation if not? 

- The Luhn Algorithm is pretty commonly used by tax authorities globally. Would it make sense to move it into `common`?

### DIFF
--- a/regimes/common/common.go
+++ b/regimes/common/common.go
@@ -53,3 +53,28 @@ func NormalizeTaxIdentity(tID *tax.Identity) error {
 	tID.Code = cbc.Code(code)
 	return nil
 }
+
+// ComputeLuhnCheckDigit expects as argument a number string excluding the check
+// digit. The returned integer should be checked against the check digit by the
+// caller.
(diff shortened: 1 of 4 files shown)


samlown (approved):
Looking good! I agree, the luhn check is probably sufficiently common to stick in a the common package.

samlown:
We base all the validation decisions on the Country Code in the `tax.Identity` object. If an `org.Party` is of a foreign entity, the tax ID's country code field should match, so we shouldn't need any more logic there. In the output to the local XML format we can adapt to whatever the schema supports.

PR #102, opened 2023-02-24
Italy fpa codes

lucanioi:
- Include list of Fattura PA Codes (`FPACodes`)  that we currently support
- Create mapping between invoice constructs to FPA Codes
- Update it/README
- Add invoice validations to italy regime
    - Validate supplier's VAT ID presence
    - Validate customer has either VAT ID or fiscal cod
    - Validate invoice meta data includes correct FPA Codes based on invoice data

### Limitations

Currently not all FPA Codes are supported. The goal is to be able to represent a typical Italian freelancer invoice using GOBL. In the beginning we will not be supporting super-specific invoice types and tax schemes. 

Using invoice meta data will enable us to move forward for now, but I don't think it's terribly scalable. Some codes, like Nature, can take on different values in multiple different parts of the final FatturaPA document. Some of these FPACodes can appear inside a list of items as well, making matters even more complicated. Currently the meta field looks like 
```
Meta["fpa-transaction-nature"] = "N5",
Meta["fpa-document-type"] = "TD15",
// etc
```
Luckily, `TipoDocumento` codes are only expected in one mandatory field at the invoice header. However, the above doesn't include information for which of the 3+ `Natura` slots `N5` should be used. Currently we'd have to do something like
```
Meta["fpa-transaction-nature-1"] = "N5",
Meta["fpa-transaction-nature-2"] = "N1.3",
// or 
Meta["fpa-some-section-transaction-nature"] = "N5",
Meta["fpa-another-section-transaction-nature"] = "N1.3",
```
We do not need to support multiple `Naturas` to generate a basic invoice. But this something to keep in mind if we want to eventually expand support for Italy.

### DIFF
--- a/regimes/it/fpa_codes.go
+++ b/regimes/it/fpa_codes.go
@@ -1,0 +1,157 @@
+package it
+
+import (
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+)
+
+// KeyDefinition defines properties of a key that is specific for a regime.
+type KeyDefinition struct {
+	// Actual key value.
+	Key cbc.Key `json:"key" jsonschema:"title=Key"`
+	// There is usually a mapping between a key and some local code.
+	Code string `json:"code,omitempty" jsonschema:"title=Code"`
+	// Short name for the key, if relevant.
+	Name i18n.String `json:"name,omitempty" jsonschema:"title=Name"`
+	// Description offering more details about when the key should be used.
+	Desc i18n.String `json:"desc,omitempty" jsonschema:"title=Description"`
+	// Any additional data that might be relevant in some regimes?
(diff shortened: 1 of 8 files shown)


regimes/it/fpa_codes.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+// FPACodeDefinition defines properties of an alphanumeric codes used by
+// FatturaPA, Italy's e-invoicing system. The codes are used to classify
+// various aspects of an invoice, namely the tax system, fund type, payment
+// method, document type, nature, and withholding type. An FPACode can include
+// uppercase letters, numbers, and a "." separator for the numeric portion.
+type FPACode string
+
+type FPACodeDefinition struct {
```
lucanioi, line 17: @samlown I wanted to make use of the new `tax.KeyDefinition` from your other PR, but I found that defining a `Key` on top of `Code` seemed a bit too much for FPA codes.

PR #105, opened 2023-03-01
Tax Tags and Scenarios

samlown:
* **Major refactor warning!!**
* Reduced Invoice Types to a small subset of mutually exclusive keys.
* `bill.InvoiceType` is now a standard `cbc.Key`
* `simplified`, `partial`, and `self-billed` invoice types have been removed and are now supported with "tax tags".
* "Schemes" have been removed and replaced with the more universal "tax tag".
* Tax Tags are simple keys with a description defined at either the document level (Regime Tags) or Tax Category Rate level.
* The new "Scenario" object allows combinations of document types and tags to be defined in order to determine if specific notes need to be added to a document (similar to the previous Scheme functionality).
* A "Scenario Summary" can be generated for each invoice in order to determine both the notes to be applied, and a combination of meta data which can be super-useful for local or format specific keys. See the Spanish examples for FacturaE.
* Now validating tax categories along side tags outside of the `Calculate` mathod.


regimes/it/categories.go
```diff
@@ -90,19 +92,24 @@ var categories = []*tax.Category{
...
 	{
-		Code:     TaxCategoryRA,
+		// IT: https://www.agenziaentrate.gov.it/portale/imposta-sul-reddito-delle-persone-fisiche-irpef-/aliquote-e-calcolo-dell-irpef
+		// EN: https://www.agenziaentrate.gov.it/portale/web/english/information-for-specific-categories-of-workers
+		Code:     TaxCategoryIRPEF,
 		Retained: true,
 		Name: i18n.String{
-			i18n.EN: "RA",
-			i18n.IT: "RA",
+			i18n.EN: "IRPEF",
```
lucanioi, line 100: Makes a lot more sense classifying the different withholding tax types separately than bunching them under RA

samlown: The names aren't obvious, but yes, there should hopefully be a proper "code" for each type of retained tax.

regimes/it/tags.go
```diff
@@ -0,0 +1,278 @@
+package it
+
+import (
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/tax"
+)
+
+// Tag keys detemoined from the "Natura" field from FatturaPA.
```
lucanioi, line 9: determined?

tax/tag.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/validation"
+)
+
+// Tag describes a tax tag that can be used to identify additional
+// requirements in an electronic invoice.
+type Tag struct {
+	// Key used to identify the tag
+	Key cbc.Key `json:"key" jsonschema:"title=Key"`
+	// Name of this scheme.
```
lucanioi, line 14: Name of this tag

tax/tag.go
```diff
@@ -0,0 +1,32 @@
...
+	Name i18n.String `json:"name,omitempty" jsonschema:"title=Name"`
+	// Human details describing what this scheme is used for.
```
lucanioi, line 16: "this tag is used for"

regimes/it/tags.go
```diff
@@ -0,0 +1,278 @@
...
+	TagKeyNotTaxable                              cbc.Key = "not-taxable" // not used since 1 Jan 2021
+	TagKeyNotTaxableExports                       cbc.Key = "not-taxable-exports"
+	TagKeyNotTaxableIntraCommunity                cbc.Key = "not-taxable-intra-community"
+	TagKeyNotTaxableSanMerino                     cbc.Key = "not-taxable-san-merino"
+	TagKeyNotTaxableExportSupplies                cbc.Key = "not-taxable-export-supplies"
+	TagKeyNotTaxableDeclarationOfIntent           cbc.Key = "not-taxable-declaration-of-intent"
+	TagKeyNotTaxableOther                         cbc.Key = "not-taxable-other"
+	TagKeyExempt                                  cbc.Key = "exempt"
+	TagKeyMarginRegime                            cbc.Key = "margin-regime"
+	TagKeyReverseCharge                           cbc.Key = "reverse-charge"
```
lucanioi, line 24: Do you see some of these moving to `common` eventually, so other regimes like `es` can specify reverse charges using tags?

samlown: Great point, but no, not in this case as the tags are very specific for Italy. The `TagKeyReverseCharge` is actually not allowed to be used any more, one of the other more descriptive codes needs to be used instead. While it is confusing, the `reverse-charge` scheme key should be used alongside the individual line tags.

lucanioi: Oh that's right, I forgot the generic `natura`s gave way to more specific versions.

lucanioi (approved):
Let's try this out!

cavalle (approved):
I can only feedback on the general approach (not the Italian specifics that I'm not familiar with). 

The changes look good to me. I suppose one of the advantages of using `tag` over `meta` is leaving `meta` for carrying information that Invopop/GOBL doesn't use. I think that reason is enough to justify the change.

I only wonder (I left a comment about it) whether we should make tags an array, rather than a single value, to make it more future-proof.

tax/set.go
```diff
@@ -23,6 +23,9 @@ type Combo struct {
 	Percent num.Percentage `json:"percent" jsonschema:"title=Percent" jsonschema_extras:"calculated=true"`
 	// Some countries require an additional surcharge (calculated if rate present).
 	Surcharge *num.Percentage `json:"surcharge,omitempty" jsonschema:"title=Surcharge" jsonschema_extras:"calculated=true"`
+	// Additional data may be required in some regimes, the tag
+	// property helps reference them.
+	Tag cbc.Key `json:"tag,omitempty" jsonschema:"title=Tag"`
```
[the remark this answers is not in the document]
samlown: I thought about this also... my conclusion was that its really complicated already and an array could open it up to abuse... my mind isn't set yet!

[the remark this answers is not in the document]
samlown: I was just reviewing the TicketBAI stuff and I think you might be right @cavalle, I'll make it an array. I've just found an example in TicketBAI where we're currently using the `meta` field and there are potentially multiple tags for each line. 🤦‍♂️

regimes/it/schemes.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/invopop/gobl/tax"
+)
+
+// Base scheme keys for this regime.
+const (
+	SchemeKeyOrdinary  cbc.Key = "ordinary"
+	SchemeKeyFreelance cbc.Key = "freelance"
+)
+
+var schemes = []*tax.Scheme{
```
lucanioi, line 16: What do you think about exporting these? In order to access available schemes from the provider, I'd have to get the `Regime` instance from Italy. Might be simpler if we can do `it.Schemes`? Same for `it.Zones` and `it.Categories`

bill/invoice.go
```diff
@@ -251,30 +253,63 @@ func (inv *Invoice) RemoveIncludedTaxes(accuracy uint32) *Invoice {
 	return &i2
 }
 
-func (inv *Invoice) prepareSchemes(r *tax.Regime) error {
+// ScenarioSummary determines a summary of the tax scenario for the invoice based on
+// the document type and tax tags.
+func (inv *Invoice) ScenarioSummary() *tax.ScenarioSummary {
```
lucanioi, line 258: I'm a fan of how this provides a much clearer/cleaner way for the client to obtain additional information than the previous iteration. I imagine it will be used like:

```go
scenarioSummary := inv.ScenarioSummary()

DatiGeneraliDocumento{
    TipoDocumento: scenarioSummary.Meta[it.KeyFatturaPATipoDocumento]
}
```

My immediate thought was it would also be helpful to have `Line.ScenarioSummary` so we can use the same technique to gain access to line-level codes for the case of Italy.

regimes/es/scenarios.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/invopop/gobl/tax"
+)
+
+var invoiceScenarios = &tax.ScenarioSet{
+	Schema: bill.ShortSchemaInvoice,
+	List: []*tax.Scenario{
+		// ** Invoice Document Types **
+		{
+			Types: []cbc.Key{bill.InvoiceTypeStandard, bill.InvoiceTypeCorrective},
+			Meta:  cbc.Meta{KeyFacturaEInvoiceDocumentType: "FC"}, // default
```
lucanioi, line 16: This is cool!

regimes/es/scenarios.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/invopop/gobl/regimes/common"
+	"github.com/invopop/gobl/tax"
+)
+
+var invoiceScenarios = &tax.ScenarioSet{
+	Schema: bill.ShortSchemaInvoice,
+	List: []*tax.Scenario{
+		// ** Invoice Document Types **
+		{
+			Types: []cbc.Key{bill.InvoiceTypeStandard, bill.InvoiceTypeCorrective},
```
lucanioi, line 15: This is cool!

tax/scenario.go
```diff
@@ -0,0 +1,114 @@
...
+			for k, v := range s.Meta {
+				summary.Meta[k] = v
+			}
+		}
+	}
+	return summary
+}
+
+func (s *Scenario) match(docType cbc.Key, docTags []cbc.Key) bool {
+	if s.Types != nil {
```
lucanioi, line 71: Would `Types` ever be `nil` given it's not a pointer type?

lucanioi: (Would it not be an empty array if not specified?)

bill/invoice.go
```diff
@@ -179,6 +184,9 @@ func (t *Totals) Validate() error {
 // Calculate performs all the calculations required for the invoice totals and taxes. If the original
 // invoice only includes partial calculations, this will figure out what's missing.
 func (inv *Invoice) Calculate() error {
+	if inv.Type == cbc.KeyEmpty {
+		inv.Type = InvoiceTypeStandard
```
lucanioi, line 188: 👍🏼 👍🏼

tax/tag.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/validation"
+)
+
+// Tag describes a tax tag that can be used to identify additional
+// requirements in an electronic invoice.
+type Tag struct {
```
lucanioi, line 11: Would it make sense for this to be in `cbc`? It's a pretty generic component and it doesn't include properties that are tax-related. It's primarily used for taxes but I could see this being useful down the line for other types

regimes/it/tags.go
```diff
@@ -0,0 +1,339 @@
...
+		},
+		Meta: cbc.Meta{
+			KeyFatturaPANatura: "N2.2",
+		},
+	},
+	{
+		Key: TagNotTaxable,
+		Name: i18n.String{
+			i18n.EN: "Not taxable (this code is no longer permitted to use on invoices emitted from 1 January 2021)",
+			i18n.IT: "Non imponibili (questo codice non è più utilizzabile a partire dal 1° gennaio 2021)",
```
lucanioi, line 143: I feel like these are descriptions rather than names. I liked how `Scheme`s had key, name, and description separately. `Key` was useful programmatically while the shorthand `Name` and a detailed `Description` provided value for end-users.

regimes/it/categories.go
```diff
@@ -90,19 +92,24 @@ var categories = []*tax.Category{
...
+			i18n.IT: "IRPEF",
 		},
 		Desc: i18n.String{
-			i18n.EN: "Withholding Tax",
-			i18n.IT: "Ritenuta d'Acconto",
+			i18n.EN: "Personal Income Tax",
+			i18n.IT: "Imposta sul Reddito delle Persone Fisiche",
+		},
+		Meta: cbc.Meta{
+			KeyFatturaPATipoRitenuta: "RT01",
```
lucanioi, line 108: I would love to be able to fetch this from the provider through a clean interface. Could we apply some variation of the "summary" pattern here?

tax/tag.go
```diff
@@ -0,0 +1,27 @@
...
+	// Key used to identify the tag
+	Key cbc.Key `json:"key" jsonschema:"title=Key"`
+	// Name of this tag.
+	Name i18n.String `json:"name,omitempty" jsonschema:"title=Name"`
+	// Additional local
+	Meta cbc.Meta `json:"meta,omitempty" jsonschema:"title=Meta"`
```
lucanioi, line 17: I liked how schemes had Key, Name, and Description. Perhaps tags can take a similar approach. Might be useful for UI? 

(I'd be happy to take that on once this gets merged!)

bill/invoice.go
samlown: Okay, so in the end I moved away from the Summaries as they technically add quite a bit of noise... the tags are at least validated on a per-line basis, but the provider is responsible for making the lookup and mapping. (We might need to review this.)

tax/scenario.go
samlown: In go, nil is the same as an empty array, but I agree, `len(s.Types) > 0` would be clearer!

tax/tag.go
samlown: I think it'll make sense as soon as we have a usage example outside of tax. Tags can abused easily, so I've mixed feelings about making them more generic.

tax/tag.go
samlown: Description could be nice! Because there are quite a few tags, I felt it was easier to have slightly longer names. Happy to refactor later though!

PR #111, opened 2023-03-22
Add other retained tax types besides IRPEF

lucanioi:
Two concerns:
- IRPEF and IRES are definitely taxes, but the rest are contributions to welfare and pension funds. They're not _quite_ taxes, but maybe I'm being too picky. 
- The "other" category doesn't quite have a code. Maybe we define `OTHER` code in `cbc` so that we can use it any time an "other" code is required?

### DIFF
--- a/regimes/it/categories.go
+++ b/regimes/it/categories.go
@@ -9,6 +9,8 @@ import (
 )
 
 // Local tax category definitions which are not considered standard.
+// There is a 6th retained tax type, RT06 "Other contributions", which is
+// currently not supported.
 const (
 	// https://www.agenziaentrate.gov.it/portale/imposta-sul-reddito-delle-persone-fisiche-irpef-/aliquote-e-calcolo-dell-irpef
 	TaxCategoryIRPEF    cbc.Code = "IRPEF"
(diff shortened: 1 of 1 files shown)


regimes/it/categories.go
```diff
@@ -11,8 +11,12 @@ import (
...
 const (
 	// https://www.agenziaentrate.gov.it/portale/imposta-sul-reddito-delle-persone-fisiche-irpef-/aliquote-e-calcolo-dell-irpef
-	TaxCategoryIRPEF cbc.Code = "IRPEF"
-	TaxCategoryINPS  cbc.Code = "INPS"
+	TaxCategoryIRPEF    cbc.Code = "IRPEF"
+	TaxCategoryIRES     cbc.Code = "IRES"
+	TaxCategoryINPS     cbc.Code = "INPS"
+	TaxCategoryENASARCO cbc.Code = "ENASARCO"
+	TaxCategoryENPAM    cbc.Code = "ENPAM"
+	TaxCategoryOther    cbc.Code = "OTHER"
```
samlown, line 19: Perhaps for the time being we can simply skip this "other" case and wait for more specific requirements to be defined, it's just simply too easy to abuse and confuse.

samlown (changes requested):
We can just hide the "other" case for the time being.

samlown (approved):
LGTM.

PR #112, opened 2023-03-23
Refactoring example handling to compile all source YAML or JSON

samlown:
* Refactor of the example handling so that all YAML or JSON files in the project will be converted to complete GOBL documents automatically.
* Objective is to make it much easier to add new examples easily.


samlown:
Force merging this as its quite useful.

PR #113, opened 2023-03-23
Italy tax id types

samlown:
* Refactor of #110
* Removing the Tax Identity "source" field.
* Adding a Tax Identity "Type"
* Moving the type validation and value definitions to the regime.
* Working on testing.


lucanioi (approved):

lucanioi:
Only thing that comes to mind is whether we need to ensure the customer has Codice Fiscale if Partita IVA is not present. As far as the FatturaPA documentation is concerned, that is the case. But I think we should just merge this and find out for real through the SDI sandbox.

cavalle (commented):
Hey! Sorry for the late feedback. 

Just a small thing I noticed is that after the merge, for some reason the new TaxIdentityType is not listed in the left sidebar TOC in https://docs.gobl.org. However, there's a link from the Tax Regime to the Tax Identity Type page. Not sure if the problem is in the generated json or in the generator itself.

Now, I think the Tax Identity Type concept is very necessary. In fact, I think we'll have to use it soon in Colombia. The case of Colombia is particular because they have two related but transversal taxonomies:

1. [Type of identifier](https://drive.google.com/open?id=1g1tsneP5GX9Trkq6zbDJK36SV4Skhku0&disco=AAAAsucPTjk) (e.g. NIT, passport, registro civil...)
2. [Type of person](https://drive.google.com/file/d/1g1tsneP5GX9Trkq6zbDJK36SV4Skhku0/edit?disco=AAAAsucPTjo) (e.g. natural person, legal person...)

At the moment, we only support invoicing for legal entities (i.e. companies) identified with an NIT, and we have hardcoded these values into the provider. This will probably have to change soon, as one of our customers needs to invoice natural persons and may need to identify them by means other than their NIT (still to be clarified). 

In any case, the question is what would be the best way to support this dual taxonomy in GOBL?

PR #114, opened 2023-03-23
Pt tax exemption

cavalle:


regimes/pt/pt.go
```diff
@@ -16,6 +17,13 @@ func init() {
 	tax.RegisterRegime(New())
 }
 
+// Custom keys used typically in meta information
+const (
+	KeyTaxCode          cbc.Key = "at-tax-code"
+	KeyTaxRegion        cbc.Key = "at-tax-country-region"
+	KeyTaxExemptionCode cbc.Key = "at-tax-exemption-code"
```
cavalle, line 24: These correspond to the the `TaxCode`, `TaxCountryRegion` and `TaxExemptionCode` fields present both in the SAFT and [webservice communication](https://info.portaldasfinancas.gov.pt/pt/apoio_contribuinte/Faturacao/Fatcorews/Documents/Comunicacao_dos_elementos_dos_documentos_de_faturacao.pdf) with the AT.

regimes/pt/pt.go
```diff
@@ -41,6 +49,23 @@ const (
 	ZoneMadeira        l10n.Code = "30" // Autonomous Region
 )
 
+// Additional zone definition for mainland Portugal
+const (
+	ZoneMainland l10n.Code = "00"
```
cavalle, line 54: Here I'm adding a new super-zone that includes all mainland PT (i.e. everything except the autonomous region). I put it in a different const section than the other zones because I'm not using an ISO-3166-2:PT but a custom code. The idea is to make it the default zone. And that way get access to the zone metadata and get its `TaxCountryCode` when needed.

Would it make sense to you if we set this zone automatically in the `calculate()` function when no other zone is given?

regimes/pt/pt.go
```diff
@@ -78,11 +104,6 @@ func New() *tax.Regime {
 								Since:   cal.NewDate(2011, 1, 1),
 								Percent: num.MakePercentage(160, 3),
 							},
-							{
-								Zones:   []l10n.Code{ZoneAzores},
-								Since:   cal.NewDate(2011, 1, 1),
-								Percent: num.MakePercentage(180, 3),
-							},
 							{
```
cavalle, line 107: This was the extraneous entry

regimes/pt/pt.go
```diff
@@ -139,6 +166,36 @@ func New() *tax.Regime {
...
+							KeyTaxCode: "ISE",
+						},
+					},
+					{
+						Key: TaxRateOther,
+						Name: i18n.String{
+							i18n.EN: "Other",
+							i18n.PT: "Outros",
+						},
+						Values: []*tax.RateValue{},
```
cavalle, line 195: No specific value is associated to this tax rate (if must be given in the percent value explicitly), so I left an empty array here. Do you think it's ok?

regimes/pt/tags.go
```diff
@@ -0,0 +1,312 @@
...
+	TagM25    cbc.Key = "m25"
+	TagM30    cbc.Key = "m30"
+	TagM31    cbc.Key = "m31"
+	TagM32    cbc.Key = "m32"
+	TagM33    cbc.Key = "m33"
+	TagM40    cbc.Key = "m40"
+	TagM41    cbc.Key = "m41"
+	TagM42    cbc.Key = "m42"
+	TagM43    cbc.Key = "m43"
+	TagM99    cbc.Key = "m99"
```
cavalle, line 38: I named the tags directly after the TaxExemptionCode they represent in Portugal. I've noticed in other countries, we use more meaningful names. In this case, though, the name for this tags isn't as obvious (at least to me) than in the other countries. Let me know if you have any suggestion for naming these tags differently

regimes/pt/pt.go
cavalle, line 24: These correspond to the `TaxCode`, `TaxCountryRegion` and `TaxExemptionCode` fields present both in the SAFT and [webservice communication](https://info.portaldasfinancas.gov.pt/pt/apoio_contribuinte/Faturacao/Fatcorews/Documents/Comunicacao_dos_elementos_dos_documentos_de_faturacao.pdf) with the AT.

regimes/pt/tags.go
lucanioi: For Italy, I used ChatGPT to come up with short names for complex tax categories, then tweaked them manually. I did wonder, though, whether using the actual codes as keys would have been clearer, and whether inventing our own internal key actually increased conceptual bloat without adding too much value. 

I guess it depends on the purpose of these keys and who will be using them primarily. Under what circumstances would a user or a developer wish the keys were given more meaningful names? 

Curious to hear what @samlown thinks about this.

regimes/pt/pt.go
lucanioi: I think `calculate()` is a good place to inject default values when absent. Do we want to distinguish between the situation where the user intentionally left it out and a case where user forgot to select the zone (in which case I guess the validation should fail)?

cavalle: Right! In this case, I think it's clear to me that the zone is not really optional. All tax reports in Portugal require to specify the `TaxCountryRegion` which possible values are (PT, PT-AZ, PT-MA). That's why I think it would make sense to make this field mandatory in PT and, for convenience, inject a "00" (i.e. Portugal Mainland) value when nothing else is provided. But perhaps we prefer users to be explicit about this?

regimes/pt/tags.go
cavalle: Exactly my thoughts while working on it. Putting myself on the shoes of a GOBL developer, I don't know if I would rather use a code that is well-known in my domain (e.g. `exempt+m01`) or some other code which only exists in the GOBL world but that may potentially be more meaningful (e.g. `exempt+outlay`). For illustration, see how our own users refer to the M99 exemption explicit by its code [here](https://invopop.slack.com/archives/C04FYMD6SLE/p1679386203186589?thread_ts=1679066241.190119&cid=C04FYMD6SLE))

Now, I've been working on this a bit more and came up with a couple of alternatives to name these tags:

Código | Tag (Legal Reference) | Tag (Meaningful)
-- | -- | --
M01 | exempt+civa+article16 | exempt+outlay
M02 | exempt+dl198-90 | exempt+intrastate-export
M04 | exempt+civa+article13 | exempt+imports
M05 | exempt+civa+article14 | exempt+exports
M06 | exempt+civa+article15 | exempt+suspension-scheme
M07 | exempt+civa+article9 | exempt+internal-operations
M09 | exempt+civa+article62b | exempt+small-retail-scheme
M10 | exempt+civa+article57 | exempt+exempt-scheme
M11 | exempt+dl346-85 | exempt+tobacco-scheme
M12 | exempt+dl221-85 | exempt+margin-scheme+travel
M13 | exempt+dl199-66+second-hand | exempt+margin-scheme+second-hand
M14 | exempt+dl199-66+art | exempt+margin-scheme+art
M15 | exempt+dl199-66+antiques | exempt+margin-scheme+antiques
M16 | exempt+riti+article14 | exempt+goods-transmission
M19 | exempt+other | exempt+other
M20 | exempt+civa+article59 | exempt+flat-rate-scheme
M21 | exempt+civa+article72 | exempt+non-deductible
M25 | exempt+civa+article38 | exempt+consignment-goods
M30 | exempt+civa+article2i | exempt+reverse-charge+waste
M31 | exempt+civa+article2j | exempt+reverse-charge+civil-eng
M32 | exempt+civa+article2l | exempt+reverse-charge+greenhouse
M33 | exempt+civa+article2m | exempt+reverse-charge+woods
M40 | exempt+civa+article6 | exempt+reverse-charge+b2b
M41 | exempt+riti+article8 | exempt+reverse-charge+intraeu
M42 | exempt+dl27-2007 | exempt+reverse-charge+real-estate
M43 | exempt+dl362-99 | exempt+reverse-charge+gold
M99 | exempt+non-taxable | exempt+non-taxable

I tend to like the last column better, because it hints you about its meaning (although, it's less precise than the legal reference or the exemption code). What do you think?

regimes/pt/pt.go
lucanioi: That makes sense to me!

regimes/pt/tags.go
lucanioi: Yea, I think you're on to something here. I'm starting to lean on the side of preferring local codes as keys and relying on `Name` to provide clarity and meaning. 

This feels like a pretty important conversation though. We should set a precedence for the future and definitely adopt a consistent approach across regimes.

samlown: I'd go with the last one. However, my intuition here is the the `M99` value is the equivalent of not defining anything. Basically, its the value I'd use in the output whenever nothing else is defined, as opposed to being an option of itself. Would that work?

samlown: Regarding Luca's comment here. Perhaps we need to clarify the intent of the longer keys. From one side you do have the "meaning" which can be transferred in a good key, but maybe thats not the only use-case, there is also the factor of error checking and typos. Its much easier to skim a document and mix up `M12` with `M21`, something which wouldn't happen with `exempt+margin-scheme+travel` and `exempt+non-deductible`.

So I would argue that its not too much of a problem to loose some meaning in the key, provided its easy to differentiate between the options available.

cavalle:
@samlown @lucanioi First of all, thanks a lot for taking the time to review the PR and leave your comments. I truly appreciate them!! ❤️

As a general response to several of your comments, I just wanted to explain that the ultimate motivation behind these changes is that both the SAFT-PT report and the communication of taxes via web service require three pieces of data to be provided:

* `TaxCode` with values `NOR` (for standard rates), `INT` (for intermediate rates), `RED` (for reduced rates), `ISE` (for exempts)  or `OUT` (for other rates, 🤷‍♂️)
* `TaxCountryRegion` with values `PT`(for continental mainland), `PT-MA` (for Madeira), `PT-AZ` (for Azores)
* `TaxExemption` with values `M01` to `M99` to denote the tax exemption when `TaxCode == "ISE"` 

Btw, these aren't InvoiceXpress-specific values or fields, this is what the AT Portugal systems expect. Now, I could perfectly map/infer those values from GOBL in the provider without making any changes here. However, I noticed that in cases that seemed similar to these in other countries (Ticketbai and Italy, notably), we had added metadata fields to the Regimes data to, I assumed, help with those mappings. 

The way I see it, this should be an all or nothing thing. That is, if I'm using metadata fields to map the GOBL zones to `PT-MA` and `PT-AZ`, I can't hard-coded in a constant in the provider for the `PT` case. I'd like to be able to map it from the same source in every case. And the same applies to the `ISE` tax code and the `M99` tax exemption which existence has been questioned in your comments.

I hope this makes sense to you! 😅 With this general comment, I'll respond to the individual notes you left (well, as soon as I can, my children are coming back home now)

regimes/pt/pt.go
cavalle: Yes, I made up the "00" code as there's no ISO-3166 code explicitly for mainland Portugal. You either have 'PT' which represents the whole country or 'PT-01'...'PT-18' for each district in mainland Portugal.

Now, I see your concerns. As I explained in my [general comment](https://github.com/invopop/gobl/pull/114#issuecomment-1483117130), my goal here was being able to map directly from a GOBL's tax zone to an AT's tax country region using metadata in the regime file (and avoid users to specify district zones in PT mainland which is fiscally irrelevant). But we don't really any of this, we can keep things simple here and do [all the mapping in the provider](https://github.com/invopop/invoicexpress/blob/main/app/domain/invoicexpress/payload_builder.rb#L142-L145). So, I'll dismiss this part of the PR.

regimes/pt/pt.go
```diff
@@ -139,6 +166,36 @@ func New() *tax.Regime {
 								Percent: num.MakePercentage(60, 3),
 							},
 						},
+						Meta: cbc.Meta{
+							KeyTaxCode: "RED",
+						},
+					},
+					{
+						Key: TaxRateExempt,
```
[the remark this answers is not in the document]
cavalle: Right! I'm totally happy with your suggestion but then, I also agree, we shouldn't use metadata for this mapping and do it in the provider. So, I'll dismiss these changes too.

regimes/pt/pt.go
cavalle: As discussed in a previous commit, we'll leave the provider deal with this.

regimes/pt/tags.go
cavalle: @samlown both `M99` and `M19` seem indeed to be catch-all categories to be used for exemptions not covered by the rest of the options but still legal. In this case, though, I tend lean towards explicitness over implicitness. Firstly, because there are two possible defaults (M99 and M19) and secondly, because the omission of an exemption tag in the GOBL may not be intentional. If we simply default to M99 when no VAT and no other exemption is given, then we're increasing the chances of integrators not realising they need to provide the right exemption reason. That's why I'll keep M99 on the list, and won't select it by default.

cavalle:
@samlown @lucanioi I just pushed the changes I mentioned in my previous comments. Namely:
* Dismissed the changes in the PT tax zones and tax rates. All the mapping data and rules will live in the provider
* Updated the tags for the tax exemptions with the meaningful keys [here](https://github.com/invopop/gobl/pull/114#discussion_r1147333384).

Comments, questions and approvals welcome!

regimes/pt/tags.go
lucanioi: Thanks for clarifying Sam, that makes sense.

lucanioi (approved):
LGTM

regimes/pt/tags.go
samlown: I don't think `M19` is a catch-all. My take is that it's a specific reason as to why the line is exempt from VAT, whereas `M99` applies when something specific is happening and there is no VAT, which is slightly different.

What I'm suggesting as a possibility is that `M99` could be applied automatically when the VAT tax combo is not there, assuming its a required field in Portugal. Happy with either approach though.

samlown (approved):
I like this simple approach. Nice one.

samlown:
@cavalle There is nothing breaking here, so I'll go ahead and merge as we're preparing a new release.

PR #115, opened 2023-03-24
Making the Tax Combo Percent property optional

samlown:
* The Tax Combo Percent is now optional, as long as a tag is provided with an explanation as to why effectively the row is "exempt" of tax.
* Related to #114


lucanioi (approved):

samlown:
> Changes make sense to me although I only one doubt about it: any tag would allow the percent to be missing, whether that tag correspond to an exemption reason or not, we don't know it. This feels a bit flaky to me. Would it make sense to allow missing percents, in principle, but then do the validation in each regime which can identify which tags aren't exceptions?

Yeah, the idea with the basic validation is just to ensure we don't have silly mistakes of neither `percent` nor `tags`. But yes, additional validation could be made on the regime level. We can work on this in other PRs. 👍

PR #116, opened 2023-03-25
Fixing cal date parsing and validation, simplify bill preceding

samlown:
* Added tests to confirm adequate `cal.Date` handling.
* Zero dates now properly handled as potentially "valid" data (fixes the JSON bidirectional parsing issue)
* Date now has its own validation method to ensure they contain reasonable data.
* `bill.Preceding`'s `issue_date` is now optional and uses a date pointer.
* Got in there a fix for the cbc note key issue in the JSON Schema


samlown:
@cavalle I'll merge this already, but you might be interested in the Preceding `issue_date` change for the stuff you were looking at with InvoiceXpress.

PR #123, opened 2023-04-05
Using cal.Period for delivery instead of custom range

samlown:
* Refactor bill delivery details to use the Period object instead of a custom range.
* Adding Period to ordering.
* cal.Period now includes a label field.


cavalle (approved):
LGTM!

PR #125, opened 2023-04-05
Add label, code, and meta to Registration

lucanioi:
Italy requires companies registered with the local Chamber of Commerce (CCIAA) to include the Numero REA (Repertorio Economico Amministrativo) and related information. 

The `NumeroREA` block in FatturaPA requires the provincial code of the CCIAA, the registration number, and a code indicating whether the company is in liquidation or not. It also requires the equity capital to be listed for certain types of companies (SpA, SApA, SRL). 

Screenshot 2023-04-05 at 17 11 55

Proposed additional fields to `org.Registration`:
- `Label`: to be used to describe the type of registration (in this case `CCIAA` or `REA`)
- `Code`: to include the registration number
- `Meta`: to include highly specific things like equity capital and liquidation status

### DIFF
--- a/org/party.go
+++ b/org/party.go
@@ -2,6 +2,8 @@ package org
 
 import (
 	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/currency"
+	"github.com/invopop/gobl/num"
 	"github.com/invopop/gobl/tax"
 	"github.com/invopop/gobl/uuid"
 
@@ -122,14 +124,16 @@ type Website struct {
 // The definition found here is based on the details required for spain.
 // If your country requires additional fields, please let us know.
 type Registration struct {
-	UUID    *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
-	Office  string     `json:"office,omitempty" jsonschema:"title=Office"`
-	Book    string     `json:"book,omitempty" jsonschema:"title=Book"`
-	Volume  string     `json:"volume,omitempty" jsonschema:"title=Volume"`
-	Sheet   string     `json:"sheet,omitempty" jsonschema:"title=Sheet"`
-	Section string     `json:"section,omitempty" jsonschema:"title=Section"`
-	Page    string     `json:"page,omitempty" jsonschema:"title=Page"`
-	Entry   string     `json:"entry,omitempty" jsonschema:"title=Entry"`
+	UUID     *uuid.UUID    `json:"uuid,omitempty" jsonschema:"title=UUID"`
+	Capital  *num.Amount   `json:"social_capital,omitempty" jsonschema:"title=Capital"`
+	Currency currency.Code `json:"currency,omitempty" jsonschema:"title=Currency"`
+	Office   string        `json:"office,omitempty" jsonschema:"title=Office"`
+	Book     string        `json:"book,omitempty" jsonschema:"title=Book"`
+	Volume   string        `json:"volume,omitempty" jsonschema:"title=Volume"`
+	Sheet    string        `json:"sheet,omitempty" jsonschema:"title=Sheet"`
+	Section  string        `json:"section,omitempty" jsonschema:"title=Section"`
+	Page     string        `json:"page,omitempty" jsonschema:"title=Page"`
+	Entry    string        `json:"entry,omitempty" jsonschema:"title=Entry"`
 }
 
 // Calculate performs any calculations required on the Party or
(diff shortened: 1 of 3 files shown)


org/party.go
lucanioi, line 1: Do we know if the existing fields are common enough to be useful in regimes other than Spain?

org/party.go
```diff
@@ -122,14 +122,20 @@ type Website struct {
...
+	Code    string `json:"code,omitempty" jsonschema:"title=Code"`
+	Office  string `json:"office,omitempty" jsonschema:"title=Office"`
+	Book    string `json:"book,omitempty" jsonschema:"title=Book"`
+	Volume  string `json:"volume,omitempty" jsonschema:"title=Volume"`
+	Sheet   string `json:"sheet,omitempty" jsonschema:"title=Sheet"`
+	Section string `json:"section,omitempty" jsonschema:"title=Section"`
+	Page    string `json:"page,omitempty" jsonschema:"title=Page"`
+	Entry   string `json:"entry,omitempty" jsonschema:"title=Entry"`
+	// Any additional useful data.
+	Meta cbc.Meta `json:"meta,omitempty" jsonschema:"title=Meta"`
```
samlown, line 138: Meta feels out of place here... meta is meant for end-users who need to communicate additional information between themselves and their customers in a given context, I don't foresee a use-case for the registration details which is very much a thing for governments as opposed to customers.

samlown (commented):
It might be useful to set up an example usage in the `/regimes/co/examples` path (using anonymous data) as its hard to imagine how this could be used from the outset.

My initial impression is that we should try to re-use fields as much as possible. A few suggestions and additions based on the screenshot:

* `office` - for the Ufficio data, even if just a two letter code
* `social_capital` - string for the amount of capital (free text is fine, no need to infer value)
*  `entry` - for the NumeroREA field (which is the registry entry)
* `tags` - this is great for handling boolean values and can use keys, so `sole-shareholder`, and `liquidating`

What do you think?

org/party.go
```diff
@@ -122,14 +123,15 @@ type Website struct {
...
-	UUID    *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
-	Office  string     `json:"office,omitempty" jsonschema:"title=Office"`
-	Book    string     `json:"book,omitempty" jsonschema:"title=Book"`
-	Volume  string     `json:"volume,omitempty" jsonschema:"title=Volume"`
-	Sheet   string     `json:"sheet,omitempty" jsonschema:"title=Sheet"`
-	Section string     `json:"section,omitempty" jsonschema:"title=Section"`
-	Page    string     `json:"page,omitempty" jsonschema:"title=Page"`
-	Entry   string     `json:"entry,omitempty" jsonschema:"title=Entry"`
+	UUID         *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
+	ShareCapital num.Amount `json:"social_capital,omitempty" jsonschema:"title=Social Capital"`
```
samlown, line 127: Go property name should match json tag. Perhaps just `capital` is enough given the context, certainly shorter which is nice. We shouldn't use a `num.Amount` for this value. There are two issues here: the currency of the document might not match, and a pointer should be used to avoid all invoices being generated with a zero value.

samlown (changes requested):
I think this simple approach makes sense. Perhaps we can just assume for the time being that we're not working with sole proprietors no companies in liquidation processes.

I can't help feeling some more tests might have caught the empty `ShareCapital` and `num.Amount` 0 value issue.

org/party.go
```diff
@@ -122,14 +124,16 @@ type Website struct {
...
-	UUID    *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
-	Office  string     `json:"office,omitempty" jsonschema:"title=Office"`
-	Book    string     `json:"book,omitempty" jsonschema:"title=Book"`
-	Volume  string     `json:"volume,omitempty" jsonschema:"title=Volume"`
-	Sheet   string     `json:"sheet,omitempty" jsonschema:"title=Sheet"`
-	Section string     `json:"section,omitempty" jsonschema:"title=Section"`
-	Page    string     `json:"page,omitempty" jsonschema:"title=Page"`
-	Entry   string     `json:"entry,omitempty" jsonschema:"title=Entry"`
+	UUID     *uuid.UUID    `json:"uuid,omitempty" jsonschema:"title=UUID"`
+	Capital  *num.Amount   `json:"social_capital,omitempty" jsonschema:"title=Capital"`
```
lucanioi, line 128: @samlown How about a `Capital` and `Currency` combination? I'm wary about leaving the `Capital` field as a simple string... FatturaPA for example require a numeric format and it would be good if we can validate it as such. 

Have you considered something like a `Money` type that always includes a currency and an amount by the way?

samlown: Yeah, I think that's okay. Saying "what's your registration capital" or "what's your registration currency", sounds reasonable.

samlown (approved):
Looks good! I'll proceed to merge so we can move this along.

PR #127, opened 2023-04-06
Add Type to bill.Charge

lucanioi:
- Add a new field `Type` to Charges
- For certain types of Charges that are distinguished from others at the regime-level. In the case of Italy, there is `Bollo`, or duty stamps, which is to be included in a special field. 

--
**Notes**
- I was going to create a `ChargeTypeDef` in the `bill` package, but it since the `bill` package depends on `tax` I was not able to utilise ChargeTypeDef in `tax.Regime`. Went with reutilising `tax.KeyDefinition` instead.

### DIFF
--- a/bill/charges.go
+++ b/bill/charges.go
@@ -35,7 +35,7 @@ func (lc *LineCharge) Validate() error {
 type Charge struct {
 	// Unique identifying for the discount entry
 	UUID *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
-	// Type of the charge, useful when regimes require special types of charges.
+	// Key for grouping or identifying charges for tax purposes.
 	Key cbc.Key `json:"key,omitempty" jsonschema:"title=Key"`
 	// Line number inside the list of discounts (calculated).
 	Index int `json:"i" jsonschema:"title=Index" jsonschema_extras:"calculated=true"`
--- a/bill/charges.go
+++ b/bill/charges.go
@@ -35,6 +35,8 @@ func (lc *LineCharge) Validate() error {
 type Charge struct {
 	// Unique identifying for the discount entry
 	UUID *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
+	// Type of the charge, useful when regimes require special types of charges.
+	Type cbc.Key `json:"key,omitempty" jsonschema:"title=Key"`
 	// Line number inside the list of discounts (calculated).
 	Index int `json:"i" jsonschema:"title=Index" jsonschema_extras:"calculated=true"`
 	// Code to used to refer to the this charge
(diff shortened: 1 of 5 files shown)


regimes/it/charges.go
```diff
@@ -0,0 +1,25 @@
+package it
+
+import (
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/tax"
+)
+
+const (
+	ChargeTypeBollo cbc.Key = "bollo"
```
samlown, line 11: It might be worth having the variable names in English.

```suggestion
	ChargeKeyStampDuty cbc.Key = "stamp-duty"
```

tax/regime.go
```diff
@@ -42,6 +42,9 @@ type Regime struct {
 	// against.
 	IdentityTypes []*IdentityType `json:"identity_types,omitempty" jsonschema:"title=Identity Types"`
 
+	// Charge types specific for the regime and may be validated or used in the UI as suggestions
+	ChargeTypes []*KeyDefinition `json:"charge_types,omitempty" jsonschema:"title=Charge Types"`
```
samlown, line 46: Also probably wants to be `ChargeKeys` instead of Types.

samlown (commented):
I think there needs to be some changes around key/type usage, so that they are more consistent.

regimes/it/charges.go
```diff
@@ -0,0 +1,26 @@
...
+// List of charge types specific to the italian regime.
+const (
+	ChargeKeyStampDuty cbc.Key = "stamp-duty"
+)
+
+var chargeKeys = []*tax.KeyDefinition{
+	{
+		Key: ChargeKeyStampDuty,
+		Name: i18n.String{
+			i18n.EN: "Stamp Stamp",
```
samlown, line 18: Typo!
```suggestion
			i18n.EN: "Stamp Duty",
```

regimes/it/charges.go
```diff
@@ -0,0 +1,26 @@
...
+			i18n.IT: "Bollo",
```
samlown, line 19: Should the name be "tassa di bollo"?

```suggestion
			i18n.IT: "Tassa di Bollo",
```

samlown (changes requested):

regimes/it/charges.go
lucanioi: yikes. fixed!

bill/charges.go
```diff
@@ -35,6 +35,8 @@ func (lc *LineCharge) Validate() error {
 type Charge struct {
 	// Unique identifying for the discount entry
 	UUID *uuid.UUID `json:"uuid,omitempty" jsonschema:"title=UUID"`
+	// Type of the charge, useful when regimes require special types of charges.
```
samlown, line 38:
```suggestion
	// Key for grouping or identifying charges for tax purposes.
```

samlown (approved):
LGTM. Just a quick suggestion there.

PR #134, opened 2023-05-03
Stamp checks

samlown:
* Validate that stamps are not duplicated.
* Only allow stamps on non-draft envelopes.

### DIFF
--- a/cbc/stamps.go
+++ b/cbc/stamps.go
@@ -1,6 +1,9 @@
 package cbc
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/invopop/validation"
 )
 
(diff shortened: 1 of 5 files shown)


cavalle (approved):
Looks good to me! 👍

One quick comment though: automatically removing stamps on calculation is the kind of behaviour that gives me mixed feelings. On the one hand, it's handy that the calculator is doing what it can to make my envelope valid. On the other hand, though, I'm sending an envelope to the system, but something else ends up being processed without any warning. If I send an envelope with stamps, it might be because there's a problem somewhere else in the system that's adding those stamps when it shouldn't. Or maybe I'm trying to process the wrong document. But since `calculate` will massage the envelope to make it valid, I won't notice any of that (at least not as early as I would if calculate didn't do anything).

Anyway, I'm not against the change. I understand that it adds value in terms of convenience, and I know that it's consistent with other behaviour already implemented in the `calculate` method. I just wanted to point out the price I think we (and our users) will be paying for it.

PR #136, opened 2023-05-04
Fixing crash when percentage rate is missing

samlown:
* Fixes bug. If for some reason `Calculate` is called on an invoice with lines that are missing a `percent` value in the included tax, the library will no longer crash.
* Leveraged this PR to include a few country name updates


cavalle (commented):
As discussed over Slack, I think we need to skip lines (or use 0% as default percent) to support the tax-exempt case.

PR #138, opened 2023-05-16
Add support for vouchers as payment method

lucanioi:
Add "voucher" payment method


samlown (commented):
What's the use-case here? These are for "payment instructions", it feels a bit strange asking someone to pay by voucher... Thinking of it this way, a few of the other methods, like "credit" make less sense also.
Would it make more sense to reflect what the "advanced" details were instead using a key?

PR #140, opened 2023-05-17
Refining payment means keys with codes, and adding to advances

samlown:
* `pay.MethodKeys` are now `pay.MeansKeys`, which is a more accurate name.
* Adding `code` property to `pay.Instructions` so that custom local codes can be used when necessary.
* Updated Italian readme docs to help better under stand the payment means.
* `pay.Advances` now also uses the list of `pay.MeansKeys`

### DIFF
--- a/pay/instructions.go
+++ b/pay/instructions.go
@@ -8,60 +8,18 @@ import (
 	"github.com/invopop/validation/is"
 )
 
-// Standard payment method codes. This is a heavily reduced list of practical
-// codes which can be linked to UNTDID 4461 counterparts.
-// If you require more payment method options, please send your pull requests.
-const (
-	MethodKeyAny            cbc.Key = "any" // Use any method available.
-	MethodKeyCard           cbc.Key = "card"
-	MethodKeyCreditTransfer cbc.Key = "credit-transfer"
-	MethodKeyDebitTransfer  cbc.Key = "debit-transfer"
-	MethodKeyCash           cbc.Key = "cash"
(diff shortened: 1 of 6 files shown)


pay/instructions.go
```diff
@@ -3,65 +3,22 @@ package pay
...
 // Instructions determine how the payment has or should be made. A
 // single "key" exists in which the preferred payment method
 // should be provided, all other details serve as a reference.
 type Instructions struct {
-	// How payment is expected or has been arranged to be collected
-	Key cbc.Key `json:"key" jsonschema:"title=Key"`
+	// The payment means expected or that have been arranged to be used to make the payment.
+	Key MeansKey `json:"key" jsonschema:"title=Key"`
+	// Additional code for the payment means if not defined in the standard list of keys.
+	Code cbc.Code `json:"code,omitempty" jsonschema:"title=Code"`
```
lucanioi, line 17: If `MeansKey` is set to `other`, then I guess we need to ensure the `Code` is present, otherwise we wouldn't know how to map it to the FatturaPA `MP` code.  Should we provide a list of available `MP`'s?

samlown: I see what you mean... the lack of a default in Italy could be confusing when set to `other`. So yes, we could add some validation there for Italy. I'll add that now. Nice catch!

samlown: Okay, validation now included to make sure if `other` is selected in Italy, the `code` is also required.

lucanioi (approved):
lgtm!

PR #141, opened 2023-05-18
Add Percentage.Rescale function

lucanioi:
- Enable rescaling of percentages 
- Previously, calling `Rescale` on a `num.Percentage` was returning a `num.Amount`

### DIFF
--- a/num/percentage.go
+++ b/num/percentage.go
@@ -80,6 +80,7 @@ func (p Percentage) StringWithoutSymbol() string {
 	return v.String()
 }
 
+// Rescale will rescale the percentage value to the provided exponent.
 func (p Percentage) Rescale(exp uint32) Percentage {
 	return Percentage{Amount: p.Amount.Rescale(exp)}
 }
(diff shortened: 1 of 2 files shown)


num/percentage.go
```diff
@@ -80,6 +80,11 @@ func (p Percentage) StringWithoutSymbol() string {
 	return v.String()
 }
 
+// Rescale will rescale the percentage value to the provided exponent.
+func (p Percentage) Rescale(exp uint32) Percentage {
+	return Percentage{Amount: p.Amount.Rescale(exp + 2)}
```
samlown, line 85: I think the `+ 2` bit might be a bit confusing. You can see the conflict in the tests:
```go
p := num.MakePercentage(160, 3)
p.String() // 16.0%
p = p.Rescale(2)
p.String() // I'd expect 16%
p = p.Rescale(4)
p.String() // I'd expect 16.00%
```

samlown (changes requested):
Comments in line, the precision stuff is a bit confusing, but I'd stick to what the `MakePercent` uses for the `Rescale` method.

num/percentage.go
lucanioi: Fixed 👍🏼

samlown (approved):
Looks Good!

PR #144, opened 2023-05-23
Fix San Merino -> San Marino

lucanioi:
Viva [San Marino](https://es.wikipedia.org/wiki/San_Marino)

### DIFF
--- a/build/regimes/it.json
+++ b/build/regimes/it.json
@@ -713,10 +713,10 @@
       }
     },
     {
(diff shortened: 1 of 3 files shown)


samlown:
If you don't mind @lucanioi, I'll merge this one manually with the #145 PR so that we don't get loads of conflicts

samlown:
Close as this is already fixed now.

PR #145, opened 2023-05-25
Rate refactor

samlown:
* Major refactor of Tax Rate handling
* Tags outside of the invoice are now gone and moved to Rates.
* `org.Item` now has it's own key field (required for TicketBAI)
* `cbc.Code` now supports adding `.`
* Moved from `cbc.Meta` to `cbc.CodeSet` for keys used potentially by regime specific conversion tools.
* Removed `Base` from `tax.CategoryTotal`


cbc/key.go
```diff
@@ -54,14 +85,42 @@ func (k Key) In(set ...Key) bool {
 	return false
 }
 
+// IsEmpty returns true if the key has no value.
+func (k Key) IsEmpty() bool {
+	return k == KeyEmpty
+}
+
+// HasValidKeyIn provides a validator to check the Key's
+// value is within the provided know set.
```
lucanioi, line 94: know set?

regimes/it/invoice_validator.go
lucanioi, line 1: what's going on here?

regimes/it/pay.go
lucanioi, line 1: Nice, I think this approach is much more consistent with how we handle other codes!!

regimes/it/scenarios.go
```diff
@@ -17,89 +80,141 @@ var invoiceScenarios = &tax.ScenarioSet{
 		// **** TIPO DOCUMENTO ****
 		{
 			Types: []cbc.Key{bill.InvoiceTypeStandard},
-			Meta: cbc.Meta{
+			Name: i18n.String{
+				i18n.EN: "Regular Invoice",
+				i18n.IT: "Fattura",
+			},
+			Codes: cbc.CodeSet{
```
lucanioi, line 87: This makes sense for Italy, I just hope it translates well to other regimes in the future.

lucanioi (approved):
Looks great, I'm looking forward to using this in the XML converter. Only thing is it seems like this doesn't include latest changes to the code (like `it/invoices.go` being replaced by `it.invoice_validator.go`

regimes/it/invoice_validator.go
samlown: I have no idea... adding this placeholder file was the only way I could get the linter to work!

regimes/pt/tags.go
```diff
@@ -7,46 +7,12 @@ import (
...
-	TagNonTaxable    cbc.Key = "non-taxable"
-)
-
 // Invoice type tags
 const (
 	TagInvoiceReceipt cbc.Key = "invoice-receipt"
 )
 
-var invoiceTags = []*tax.Tag{
+var invoiceTags = []*tax.KeyDefinition{
```
cavalle, line 15: @samlown I noticed that, in IT, you moved the `invoiceTags` to `scenarios.go` and deleted `tags.go`. Do we want to do the same in PT?

samlown: yes, that does make sense... moving.

PR #147, opened 2023-05-26
Fix combining rates when the keys are different

samlown:
* Fixes an issue in Italy whereby multiple income tax rate definitions in the same invoice were being combined into one.


lucanioi (approved):
LGTM!

PR #150, opened 2023-05-29
Add TaxCountryRegion & TaxCode codes to PT regime

cavalle:
* Adds `TaxCountryRegion` and `TaxCode` codes to the PT regime
* Adds table documentation on how to map GOBL fields to those codes
* Enables replacing hard-coded mapping logic in IX provider (see https://github.com/invopop/invoicexpress/pull/35)

### DIFF
--- a/regimes/pt/README.md
+++ b/regimes/pt/README.md
@@ -21,9 +21,43 @@ Portugal doesn't have an e-invoicing format per se. Tax information is reported 
 | ND | Nota de débito | `credit-note` | |
 | NC | Nota de crédito | `debit-note` | |
 
+### `TaxCountryRegion` (País ou região do imposto)
+
+| Code | Name | GOBL Tax Identity Zone |
+| --- | --- | --- |
+| PT | Aveiro | `01` |
+| PT | Beja | `02` |
+| PT | Braga | `03` |
+| PT | Bragança | `04` |
+| PT | Castelo Branco | `05` |
+| PT | Coimbra | `06` |
+| PT | Évora | `07` |
+| PT | Faro | `08` |
+| PT | Guarda | `09` |
+| PT | Leiria | `10` |
+| PT | Lisboa | `11` |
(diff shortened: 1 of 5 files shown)


regimes/pt/README.md
```diff
@@ -21,9 +21,43 @@ Portugal doesn't have an e-invoicing format per se. Tax information is reported
 | ND | Nota de débito | `credit-note` | |
 | NC | Nota de crédito | `debit-note` | |
 
+### `TaxCountryRegion` (País ou região do imposto)
+
```
samlown, line 25: I think we need more details here for how these fields will be used, I tried to do the same for Italy already. The "GOBL Tax Identity Zone" also feels a bit out of place, it looks like we made up a number. These are actually the zone codes from ISO 3166-2:PT (according to the comment in the code: https://github.com/invopop/gobl/blob/main/regimes/pt/pt.go#L23)

regimes/pt/tax_categories.go
```diff
@@ -135,6 +144,7 @@ var taxCategories = []*tax.Category{
 					i18n.PT: "Artigo 16.°, n.° 6 do CIVA",
 				},
 				Codes: cbc.CodeSet{
+					KeyATTaxCode:          "ISE",
```
samlown, line 147: I'm surprised the linter didn't pick up on this... might be better to have constants when the code is repeated so much?

regimes/pt/README.md
cavalle: Local codes descriptions added in f055a5bde1a0e2f7dcdff3ab3ffaea1bd95e54af. Thank you for the feedback!

regimes/pt/tax_categories.go
cavalle: Repeated codes extracted to constants in 82b32dbb5066a08dc8417aaf22532d6edbe4608e. Thank you for the feedback!

PR #151, opened 2023-05-29
Passing the draft status of an envelope to validation

samlown:
* Documents can now check if the envelope is in draft mode by looking at the context.
* Useful for allowing the Invoice Code field to be blank while in draft, but not allowed when transitioning to a closed invoice, which is the default if no envelope is present.


cavalle (approved):
LGTM!

PR #154, opened 2023-06-15
WIP Mexico Validations

cavalle:


regimes/mx/invoice_validator.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (v *invoiceValidator) validate() error {
+	inv := v.inv
+	return validation.ValidateStruct(inv,
+		validation.Field(&inv.Customer,
+			validation.Required,
+			validation.By(v.validCustomer),
+		),
+		validation.Field(&inv.Lines,
```
samlown, line 26: This is the line that causes each `Line` to be validated and thus causes the regime error due to the missing context! The `ValidateStruct` call above would need to be a `ValidateStructWithContext`, whose context comes from an earlier `ValidateWithContext` call.

To get around this, there are a few options.
 * Move the `bill.Line` validation to the main `mx.Validate(obj interface{})` method so that it won't be re-validating the line, or,
 * Add support for the context in the encompassing `validate()` methods.

PR #155, opened 2023-06-16
Validations for bare minimum MX invoice

cavalle:
* Adds validations necessary to generate an invoice with the bare minimum amount of attributes in Mexico
* Fixes a crash when a line without an item was being calculated

### DIFF
--- a/regimes/mx/invoice_validator.go
+++ b/regimes/mx/invoice_validator.go
@@ -1,0 +1,72 @@
+package mx
+
+import (
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/validation"
+)
+
+type invoiceValidator struct {
+	inv *bill.Invoice
+}
+
+func validateInvoice(inv *bill.Invoice) error {
+	v := &invoiceValidator{inv: inv}
+	return v.validate()
+}
+
+func (v *invoiceValidator) validate() error {
+	inv := v.inv
+	return validation.ValidateStruct(inv,
+		validation.Field(&inv.Customer,
+			validation.Required,
+			validation.By(v.validCustomer),
+		),
+		validation.Field(&inv.Lines,
+			validation.Each(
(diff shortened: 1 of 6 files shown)


regimes/mx/invoice_validator_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/invopop/gobl/cal"
+	"github.com/invopop/gobl/l10n"
+	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/gobl/tax"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func validInvoice(t *testing.T) *bill.Invoice {
```
samlown: you can add a `t.Helper()` call at the start of the method to avoid the linter unused param issue.

regimes/mx/invoice_validator.go
```diff
@@ -0,0 +1,72 @@
...
+		validation.Field(&line.Quantity, validation.By(v.validatePositiveAmount)),
+		validation.Field(&line.Total, validation.By(v.validatePositiveAmount)),
+		validation.Field(&line.Taxes,
+			validation.Required,
+			validation.Skip, // Prevents each tax's `ValidateWithContext` function from being called again.
+		),
+	)
+}
+
+func (v *invoiceValidator) validatePositiveAmount(value interface{}) error {
```
samlown, line 62: For sure these methods can be made more general in the `num` package! I'll make a PR with suggestions :-)

samlown (approved):
LGTM!

regimes/mx/invoice_validator_test.go
cavalle: I thought that the linter was right that I wasn't really using it, so I just removed it. But the trick is good to know

regimes/mx/invoice_validator.go
cavalle: That'll be great. Thanks a lot!

PR #156, opened 2023-06-16
Adding validation rules for min or max amounts

samlown:
* Amounts and percentages can now be validated using `num.Min()` and `num.Max()` rules.

### DIFF
--- a/num/validation.go
+++ b/num/validation.go
@@ -26,9 +26,9 @@ const (
 
 var (
 	// Positive validates the that value is greater than or equal to zero.
-	Positive = Min(MakeAmount(0, 0))
+	Positive = Min(MakeAmount(0, 0)).Exclusive()
 	// Negative validates the value is less than or equal to zero.
-	Negative = Max(MakeAmount(0, 0))
+	Negative = Max(MakeAmount(0, 0)).Exclusive()
 	// NotZero validates that the value is not zero.
 	NotZero = ThresholdRule{
 		threshold: Amount{0, 0},
--- a/num/validation.go
+++ b/num/validation.go
@@ -63,6 +63,8 @@ func (r ThresholdRule) Exclusive() ThresholdRule {
 	return r
 }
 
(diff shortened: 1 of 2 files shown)


cavalle (approved):
Nice! This will definitely simplify the MX validation [here](https://github.com/invopop/gobl/pull/157/files#diff-ec7d365e0651975b24ecda758c0ee044b04cab6851ed9fb82897201df1ec7d17). Thinking about that case I was wondering, for the Ultimate Convenience™️, we could also add, on top of those `num.Min` and `num.Max`, these additional helpers: `num.Positive`, `num.Zero` and `num.Negative`. WDYT?

samlown:
> Nice! This will definitely simplify the MX validation [here](https://github.com/invopop/gobl/pull/157/files#diff-ec7d365e0651975b24ecda758c0ee044b04cab6851ed9fb82897201df1ec7d17). Thinking about that case I was wondering, for the Ultimate Convenience™️, we could also add, on top of those `num.Min` and `num.Max`, these additional helpers: `num.Positive`, `num.Zero` and `num.Negative`. WDYT?

Yeah, I'll add a couple of helper methods :-) Not sure why you'd want to check a number was zero though... 😆

cavalle (commented):
@samlown thank you for considering my feedback. Shouldn't `Positive` and `Negative` be exclusive? Zero is neither positive nor negative

samlown:
> @samlown thank you for considering my feedback. Shouldn't `Positive` and `Negative` be exclusive? Zero is neither positive nor negative

So my initial idea was to consider 0 as simply an empty value and thus ignored by the Positive and Negative rules, and simply use the `NonZero` rule in combination if required.

However, you're right that this isn't strictly semantically correct, zero is neither positive nor negative. I've made the change to match your suggestion. I hope that users who need these rules will use a `nil` value in the case of there not being a value as opposed to using `0.0`.

cavalle (approved):
Looks great!

PR #157, opened 2023-06-19
Refactor to use num.Min validation

cavalle:
* Nice simplification made possible by https://github.com/invopop/gobl/pull/156


samlown (commented):
Nice.

PR #158, opened 2023-06-26
`FormaPago` and `UsoCFDI` mapping codes for MX

cavalle:
* Adds mapping codes from GOBL's payment instructions keys (i.e. payment means) to the CFDI's `FormaPago` field
* Adds mapping codes from GOBL's invoice tax tags (and scenarios) to the CFDI's `UsoCFDI` field
* Extracts and makes public a `TaxRegime` method that will be used by [the related changes](https://github.com/invopop/gobl.cfdi/pull/6) in `gobl.cfdi`
* Fixes a typo in PT's a README

### DIFF
--- a/regimes/mx/README.md
+++ b/regimes/mx/README.md
@@ -62,7 +62,7 @@ The CFDI’s `FormaPago` field specifies an invoice's means of payment. The foll
 | 24 | Confusión | `other+merger` |
 | 25 | Remisión de deuda | `other+remission` |
 | 26 | Prescripción o caducidad | `other+expiration` |
-| 27 | A satisfacción del acreedor | `other+extinguishment` |
+| 27 | A satisfacción del acreedor | `other+satisfy-creditor` |
 | 28 | Tarjeta de débito | `card+debit` |
 | 29 | Tarjeta de servicios | `card+services` |
 | 30 | Aplicación de anticipos | `other+advance` |
--- a/regimes/mx/README.md
+++ b/regimes/mx/README.md
@@ -5,3 +5,67 @@ Mexico uses the CFDI (Comprobante Fiscal Digital por Internet) format for their 
 ## Public Documentation
 
 * [Formato de factura (Anexo 20)](http://omawww.sat.gob.mx/tramitesyservicios/Paginas/anexo_20.htm)
+
+
+## Local Codes
+
+### `UsoCFDI`
+
+The CFDI’s `UsoCFDI` field specifies how the invoice's recipient will use the invoice to deduce taxes for the expenditure made. The following table list all the supported values and how GOBL will map them from the invoice's tax tags:
+
+| Code | Name | GOBL Tax Tag |
+| --- | --- | --- |
+| G01 | Adquisición de mercancías | `use+goods-acquisition` |
+| G02 | Devoluciones, descuentos o bonificaciones | `use+returns` |
+| G03 | Gastos en general | `use+general-expenses` |
+| I01 | Construcciones | `use+construction` |
+| I02 | Mobiliario y equipo de oficina por inversiones | `use+office-equipment` |
+| I03 | Equipo de transporte | `use+transport-equipment` |
+| I04 | Equipo de computo y accesorios | `use+computer-equipment` |
+| I05 | Dados, troqueles, moldes, matrices y herramental | `use+manufacturing-tooling` |
+| I06 | Comunicaciones telefónicas | `use+telephone-comms` |
+| I07 | Comunicaciones satelitales | `use+satellite-comms` |
+| I08 | Otra maquinaria y equipo | `use+other-machinery` |
+| D01 | Honorarios médicos, dentales y gastos hospitalarios | `use+medical-expenses` |
+| D02 | Gastos médicos por incapacidad o discapacidad | `use+medical-expenses+disability` |
+| D03 | Gastos funerales | `use+funeral-expenses` |
+| D04 | Donativos | `use+donation` |
+| D05 | Intereses reales efectivamente pagados por créditos hipotecarios (casa habitación) | `use+mortgage-interest` |
+| D06 | Aportaciones voluntarias al SAR | `use+sar-contribution` |
+| D07 | Primas por seguros de gastos médicos | `use+medical-insurance` |
+| D08 | Gastos de transportación escolar obligatoria | `use+school-transportation` |
+| D09 | Depósitos en cuentas para el ahorro, primas que tengan como base planes de pensiones | `use+savings-deposit` |
+| D10 | Pagos por servicios educativos (colegiaturas) | `use+school-fees` |
+| S01 | Sin efectos fiscales | `use+no-tax-effects` |
+| CP01 | Pagos | `use+suplementary-payment` |
+| CN01 | Nómina | `use+payroll` |
+
+### `FormaPago`
+
+The CFDI’s `FormaPago` field specifies an invoice's means of payment. The following list list all the supported values and how GOBL will map them from the invoice's payment instructions key:
+
+| Code | Name | GOBL Payment Instructions Key |
+| --- | --- | --- |
+| 01 | Efectivo | `cash` |
+| 02 | Cheque nominativo | `cheque` |
+| 03 | Transferencia electrónica de fondos | `credit-transfer` |
+| 04 | Tarjeta de crédito | `card` |
+| 05 | Monedero electrónico | `online+wallet` |
+| 06 | Dinero electrónico | `online` |
+| 08 | Vales de despensa | `other+food-vouchers` |
+| 12 | Dación en pago | `other+in-kind` |
+| 13 | Pago por subrogación | `other+subrogation` |
+| 14 | Pago por consignación | `other+consignation` |
+| 15 | Condonación | `other+condonation` |
+| 17 | Compensación | `netting` |
+| 23 | Novación | `other+novation` |
+| 24 | Confusión | `other+conflicting` |
+| 25 | Remisión de deuda | `other+remission` |
+| 26 | Prescripción o caducidad | `other+expiration` |
+| 27 | A satisfacción del acreedor | `other+extinguishment` |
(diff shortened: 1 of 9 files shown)


regimes/mx/invoice_validator.go
```diff
@@ -30,9 +37,23 @@ func (v *invoiceValidator) validate() error {
...
+			validation.Required,
+			validation.By(v.validPayment),
+		),
 	)
 }
 
+func (v *invoiceValidator) validateScenarios() error {
+	ss := v.inv.ScenarioSummary()
+
+	if ss.Codes[KeySATUsoCFDI] == "" {
```
samlown, line 50: This feels a bit strange, especially the resulting error message as its not obvious what is missing. Perhaps the objective here is just to make sure that the invoice has a matching scenario and we can validate that instead?

regimes/mx/scenarios.go
```diff
@@ -0,0 +1,388 @@
...
+			i18n.ES: "Nómina",
+		},
+	},
+}
+
+var scenarios = []*tax.ScenarioSet{
+	invoiceScenarios,
+}
+
+var invoiceScenarios = &tax.ScenarioSet{
```
samlown, line 216: Is there an opportunity for a default scenario? like "General Expenses"? We could be both explicit and have a default.

samlown (commented):
Code is obviously great, but a few things to potentially discuss, especially with the scenarios.

cavalle:
@samlown thanks a lot for the thorough review! I'll respond to each comment

regimes/mx/README.md
```diff
@@ -5,3 +5,67 @@ Mexico uses the CFDI (Comprobante Fiscal Digital por Internet) format for their
...
+| 12 | Dación en pago | `other+in-kind` |
+| 13 | Pago por subrogación | `other+subrogation` |
+| 14 | Pago por consignación | `other+consignation` |
+| 15 | Condonación | `other+condonation` |
+| 17 | Compensación | `netting` |
+| 23 | Novación | `other+novation` |
+| 24 | Confusión | `other+conflicting` |
+| 25 | Remisión de deuda | `other+remission` |
+| 26 | Prescripción o caducidad | `other+expiration` |
+| 27 | A satisfacción del acreedor | `other+extinguishment` |
```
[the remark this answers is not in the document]
cavalle: For what I understand, this not a promise of a future payment (i.e. promissory note) but a written agreement between the debtor and the creditor to "extinguish" the obligations of the latter. See for instance this explanation from [here](https://facturama.mx/blog/que-significa/a-satisfaccion-del-acreedor/):

"La Forma de Pago 27 es el acuerdo firmado entre deudor y acreedor en el que ambos ratifican su común decisión de extinguir las obligaciones (que son de interés del acreedor)."

I picked "extinguishment" based on that definition, but I wasn't super-happy with the choice either. It's a tricky one! WDYT?

regimes/mx/invoice_validator.go
cavalle: Yes, this one was a bit tricky. 

I borrowed the idea of validating scenarios from [the implementation in Italy](https://github.com/invopop/gobl/blob/main/regimes/it/invoice_validator.go#L125-L139). There, the error messages are like "missing scenario related to TipoDocumento" which I found that a GOBL user may find hard to decipher as scenarios are an internal, undocumented concept they don't need to be familiar with.

So I tried to come up with an error message that tells the users what they need do to fix the GOBL document using only concepts they can understand. Maybe a better wording would have been: "invoice must include a tax tag that maps to a UsoCFDI code". WDYT?

regimes/mx/README.md
```diff
@@ -5,3 +5,67 @@ Mexico uses the CFDI (Comprobante Fiscal Digital por Internet) format for their
...
+| 02 | Cheque nominativo | `cheque` |
+| 03 | Transferencia electrónica de fondos | `credit-transfer` |
+| 04 | Tarjeta de crédito | `card` |
+| 05 | Monedero electrónico | `online+wallet` |
+| 06 | Dinero electrónico | `online` |
+| 08 | Vales de despensa | `other+food-vouchers` |
+| 12 | Dación en pago | `other+in-kind` |
+| 13 | Pago por subrogación | `other+subrogation` |
+| 14 | Pago por consignación | `other+consignation` |
+| 15 | Condonación | `other+condonation` |
```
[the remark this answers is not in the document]
cavalle: Fixed on https://github.com/invopop/gobl/pull/158/commits/91645bee6271d03cd7131453e84b8afb675d9af9

regimes/mx/README.md
```diff
@@ -5,3 +5,67 @@ Mexico uses the CFDI (Comprobante Fiscal Digital por Internet) format for their
...
+
+| Code | Name | GOBL Payment Instructions Key |
+| --- | --- | --- |
+| 01 | Efectivo | `cash` |
+| 02 | Cheque nominativo | `cheque` |
+| 03 | Transferencia electrónica de fondos | `credit-transfer` |
+| 04 | Tarjeta de crédito | `card` |
+| 05 | Monedero electrónico | `online+wallet` |
+| 06 | Dinero electrónico | `online` |
+| 08 | Vales de despensa | `other+food-vouchers` |
```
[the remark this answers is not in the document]
cavalle: Fixed on https://github.com/invopop/gobl/pull/158/commits/91645bee6271d03cd7131453e84b8afb675d9af9

regimes/mx/README.md
```diff
@@ -5,3 +5,67 @@ Mexico uses the CFDI (Comprobante Fiscal Digital por Internet) format for their
...
+| 12 | Dación en pago | `other+in-kind` |
+| 13 | Pago por subrogación | `other+subrogation` |
+| 14 | Pago por consignación | `other+consignation` |
```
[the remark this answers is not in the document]
cavalle: Fixed on https://github.com/invopop/gobl/pull/158/commits/91645bee6271d03cd7131453e84b8afb675d9af9

regimes/mx/README.md
```diff
@@ -5,3 +5,67 @@ Mexico uses the CFDI (Comprobante Fiscal Digital por Internet) format for their
...
+| 15 | Condonación | `other+condonation` |
+| 17 | Compensación | `netting` |
+| 23 | Novación | `other+novation` |
+| 24 | Confusión | `other+conflicting` |
```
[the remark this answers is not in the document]
cavalle: Fixed on https://github.com/invopop/gobl/pull/158/commits/91645bee6271d03cd7131453e84b8afb675d9af9

regimes/mx/README.md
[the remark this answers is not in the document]
samlown: hmm 🤔 

Maybe something more explicit?

```suggestion
| 27 | A satisfacción del acreedor | `other+satisfy-creditor` |
```

regimes/mx/scenarios.go
cavalle: I always got mixed feelings about defaults. Particularly when they are implicit (i.e. invisible). While they may be convenient, my concern is the user may forget to override the default scenario (no validation error will remind them about it) when that's necessary. 

In other regimes, we're using scenarios mostly for setting subtypes of documents. In that case, there's a natural hierarchy of subtypes and it makes sense to have a default scenario associated with the default invoice type (which is calculated if missing) and no tags. That is not the case here, though.

That's my point of view, but I trust your instinct better than mine on this point. Let me know what you think.

regimes/mx/README.md
```diff
@@ -5,3 +5,67 @@ Mexico uses the CFDI (Comprobante Fiscal Digital por Internet) format for their
...
+| 25 | Remisión de deuda | `other+remission` |
+| 26 | Prescripción o caducidad | `other+expiration` |
+| 27 | A satisfacción del acreedor | `other+extinguishment` |
+| 28 | Tarjeta de débito | `card+debit` |
+| 29 | Tarjeta de servicios | `card+services` |
+| 30 | Aplicación de anticipos | `other+advance` |
+| 31 | Intermediario pagos | `other+intermediary` |
+| 99 | Por definir | `other+tbd` |
```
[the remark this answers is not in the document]
samlown: I still think just "other" is good as a catch-all 👍 it does make me wonder however is `other` should be combined... in theory there should be a solid enough base in GOBL to reflect most forms of payment or contract

regimes/mx/scenarios.go
samlown: Its a tough one! There is a strong argument for saying that GOBL deals with the most common use-case: Digital goods. We can chat about this later if you have time.

regimes/mx/invoice_validator.go
samlown: Perhaps the message should just be `"'use' tax tag is required"`? But I agree the Italian example isn't very clear either, whatever we decide here should be reapplied there.

regimes/mx/README.md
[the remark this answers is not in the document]
cavalle: Solved in https://github.com/invopop/gobl/pull/158/commits/fad21e4f650ce7788b16c771813ca3a5763d44e1

regimes/mx/invoice_validator.go
cavalle: Solved in https://github.com/invopop/gobl/pull/158/commits/c44f03520ad0bd823b058b0abd9a61cea8502231

cavalle: Although, in the case of the Italian cases, I don't think we can do like here. Both the invoice type and the tax tags tags may be very well present and the combination of them not map to any scenario. So, the best I think we could do there is something along the lines of my previous suggestion: "invoice type and tax tags doesn't map to a TipoDocumento code"

regimes/mx/scenarios.go
cavalle: @samlown I'm merging this PR to move forward with other changes that depend on these. But whether we want to set a default here or not is still open to discussion. I'm taking a note to ensure we chat about it whenever we have a chance.

PR #160, opened 2023-06-28
Support for Bill Correction Options object

samlown:
* Support providing a pre-filled object for options sent to correct an invoice.


cavalle (approved):
LGTM!

PR #167, opened 2023-07-07
Additional validations for basic fields in MX

cavalle:
* Validates payment terms max length (for mapping to `CondicionesDePago`)
* Validates item units presence and adds a few commonly used units (for mapping to `ClaveUnidad`)
* Validates custom "SAT" item identity presence and format (for mapping to `ClaveProdServ`)
* Fixes a couple of typos in the README
* Related to: https://github.com/invopop/gobl.cfdi/pull/8


regimes/mx/README.md
```diff
@@ -101,6 +101,66 @@ The following GOBL maps to the `05` (Monedero electrónico) value of the `FormaP
...
+  "$schema": "https://gobl.org/draft-0/bill/invoice",
+
+  // [...]
+
+  "lines": [
+    {
+      // [...]
+
+      "item": {
+        "name": "Alive ducks",
```
samlown, line 155: Translation would be "Live ducks" 🦆😆 

```suggestion
        "name": "Live ducks",
```

regimes/mx/invoice_validator.go
```diff
@@ -10,6 +11,9 @@ import (
 	"github.com/invopop/validation"
 )
 
+// SAT item identity codes (ClaveProdServ) regular expression.
+var itemIdentityCodeRegexp = regexp.MustCompile(`^\d{8}$`)
```
samlown, line 15: Is it worth allowing 6 digits and assuming the last two are `00`? Probably not... just curious :-)

regimes/mx/invoice_validator.go
```diff
@@ -77,9 +81,37 @@ func (v *invoiceValidator) validLine(value interface{}) error {
 			validation.Required,
 			validation.Skip, // Prevents each tax's `ValidateWithContext` function from being called again.
 		),
+		validation.Field(&line.Item, validation.By(v.validItem)),
 	)
 }
 
+func (v *invoiceValidator) validItem(value interface{}) error {
```
samlown, line 88: Rather than just validating inside the context of an invoice, perhaps they could be validated in general. I can imagine a scenario whereby a user has a database of `org.Item` to be used inside invoices. It makes sense that they should be validatable in the context of a tax regime as independent objects. (Let me know if you'd like to me to and suggest the change in a PR into this one)

samlown (commented):
A few discussion points to clarify...

samlown:
@cavalle please see #168

cavalle:
> @cavalle please see #168

That's great! I'll refactor the MX validations as soon as the changes are merged

regimes/mx/invoice_validator.go
cavalle: Sure! I suppose this comment was your motivation for the changes you added in #168. Happy to refactor to validate directly rather than in the context of the invoice as soon as the changes are merged!

regimes/mx/invoice_validator.go
cavalle: We can do it! Would you do it as a kind of "normalisation" of the code and add the missing 00 when on document calculation?

regimes/mx/README.md
cavalle: 🤦 Thanks for the correction!

org/unit.go
```diff
@@ -38,6 +39,9 @@ const (
 	UnitHour         Unit = `h`
 	UnitMinute       Unit = `min`
 	UnitPiece        Unit = `piece`
+	UnitEach         Unit = `each`
```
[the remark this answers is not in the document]
cavalle: "Each" is indeed weird. I used simply because that's the name that appears in the UNECE table and is short:

image

But, on a second thought, in the MX catalogue, they call it "Elemento" and I've notice that "Item" is not used for anything else in UNECE. "Item" is also short and, although a bit too generic, definitely less strange than "Each". 

I was also thinking that perhaps we could make "Service" and "Activity" shorter: `serv` and `act`. WDYT?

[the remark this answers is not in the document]
samlown: Okay, I'm included to agree that `item` (a noun) has more meaning than `each` (a determiner or pronoun). I was going to suggest that "each" could be understood for both services and products, but then the description says "item", which rules that idea out.

Think about this... it would be nice to have the official descriptions there too. Might be worth a quick copy and paste session there.

I think the `service` and `activity` names will be used so infrequently that we can keep the long versions for clarity.

regimes/mx/invoice_validator.go
samlown: That could indeed make sense! Might be hard to get the normalization to the item identity level though. I think just validating either 6 or 8 would be sufficient.

regimes/mx/invoice_validator.go
samlown: Cool! That's now merged!

org/unit.go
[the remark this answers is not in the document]
cavalle: Feedback address in 04d35d6

org/unit.go
```diff
@@ -14,7 +14,8 @@ type Unit cbc.Key
 // many different unit codes in the world, that it's impractical to try and define them
 // all, this is thus a selection of which we think are the most useful.
 const (
-	UnitEmpty Unit = `` // No unit defined
+	UnitEmpty  Unit = ``       // No unit defined
+	UnitMutual Unit = `mutual` // Unit agreed between the parties
```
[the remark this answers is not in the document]
cavalle: Addressed in 04d35d6 as I explained in my previous comment

regimes/mx/invoice_validator.go
cavalle: Addressed in 7a6ab005bc1640570fa51ac6a1868ad7ff2fb077

regimes/mx/invoice_validator.go
cavalle: Cool! Since that padding of zeroes should happen either here or in `gobl.cfdi` anyway (i.e., the SAT doesn't accept 6-digit codes), I gave it a try to do it here and I think the resulting code is good enough (particularly after the refactor that #168 enabled). See e825e46, and let me know if you have any feedback.

cavalle:
@samlown After the latest changes I added based on your feedback, I'm going to merge this PR to let our clients play with them. But you may still want to review those late changes, and I'll be more than happy to consider any additional feedback you may have.

regimes/mx/invoice_validator.go
samlown: I wasn't sure if that would work! Nice one!

PR #166, opened 2023-07-05
Improving correction error messages

samlown:
* Correction errors improved.
* Adding option to embed build assets in binaries.

PR #168, opened 2023-07-10
Regime validation and better corrective messages

samlown:
Replaces #166.

* Improving correction error messages
* Fixing validation bug for Party Addresses.
* Added `tax.ValidateStructWithRegime` method to make it easy to validate more types of object inside the context of a regime.

### DIFF
--- a/bill/invoice_correct.go
+++ b/bill/invoice_correct.go
@@ -169,7 +169,7 @@ func (inv *Invoice) Correct(opts ...cbc.Option) error {
 			inv.Type = InvoiceTypeCorrective
 			inv.Invert()
 		} else {
-			return errors.New("credit not supported by regime")
+			return errors.New("credit note not supported by regime")
 		}
 		inv.Payment.ResetAdvances()
 	} else if o.Debit {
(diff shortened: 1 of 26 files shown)


cavalle (approved):

samlown:
> LGTM! And it's great to have more structs available to be validated directly by its type. Now, I'm just curious about what's the criteria to allow certain structs to be validated in the regime and others not. Put otherwise, what's the reason not to use the new `tax.ValidateStructWithRegime` in every struct?

Agreed. There should be a criteria. I think the criteria I used without realising is that all "exposed" structs should be validatable within a regime. So from "pay/pay.go" file, `Advance`, `Instructions`, and `Terms` should be validatable with a regime, but nothing else. I'll make some changes to the PR.

cavalle:
> Agreed. There should be a criteria. I think the criteria I used without realising is that all "exposed" structs should be validatable within a regime. So from "pay/pay.go" file, `Advance`, `Instructions`, and `Terms` should be validatable with a regime, but nothing else. I'll make some changes to the PR.

That criteria is spot on! 🎯

PR #172, opened 2023-07-19
Assign currency and update numbers

samlown:
* Ensuring that all source entered data has the minimum precision of the currency in use.
* Refactoring invoice total calculations with more tests.

### DIFF
--- a/bill/charges.go
+++ b/bill/charges.go
@@ -92,3 +92,23 @@ func (m *Charge) removeIncludedTaxes(cat cbc.Code, accuracy uint32) *Charge {
 	m2.Amount = m2.Amount.Upscale(accuracy).Remove(*rate.Percent)
 	return &m2
 }
+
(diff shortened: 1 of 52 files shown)


cavalle (approved):
Changes LGTM and I'm very much in favour of this "normalisation" of amounts to have, at least, the precision of the currency. This will relieve GOBL producers from the responsibility of appending trailing zeroes to their numbers when they don't have the accuracy of the currency.

I only got a contribution regarding the implementation: I'm not sure I would expect the functions that calculate the totals of charges, outlays, discounts or lines to be also responsible from calculating those structs themselves. To illustrate with an example, I wouldn't expect that a calling a function like `lineTotal` mutates each line by setting the index, the percent of the taxes, change the precision of amounts… I _think_ I would decouple the calculation of each struct from the calculation of the totals of that struct.

cavalle:
The (perhaps) conflated responsibilities are still there, but the new names are a clear improvement. We can always refactor out in the future. Thanks for considering my feedback!

PR #173, opened 2023-07-21
Rename Calculate to Normalize

samlown:
* Quickly experimenting renaming Calculate into Normalize which more accurately describes the intent.


samlown:
"Calculate" is fine for the time being.

PR #174, opened 2023-07-24
Save example output and compare

samlown:
* This makes the example documents easier to test against.
* Any differences between the existing generated envelope and a new one, will cause an error that a human will need to review and run the tests with the `--update` flag.
* Hopefully this will make it easier to detect any unexpected consequences of changes.


cavalle (approved):
Looks great to me! Thank for including the MX example

PR #175, opened 2023-07-26
Credit notes (and stamps) to MX

cavalle:
* Adds SAT UUID Stamp key to be linked from preceding credit notes
* Uses scenarios to map the invoice type to the fields `TipoRelacion` and `TipoDeComprobante`
* Adds preceding definitions to produce credit notes from invoices
* Implements related validations


samlown (approved):
Looks good!

PR #176, opened 2023-07-26
Determine precision using trial and error loop

samlown:
* Rather than trying calculate the precision required, we just trial and error with incremental accuracies until one fits.

### DIFF
--- a/bill/invoice.go
+++ b/bill/invoice.go
@@ -209,42 +209,60 @@ func (inv *Invoice) Calculate() error {
 // RemoveIncludedTaxes is a special function that will go through all prices which may include
 // the tax included in the invoice, and remove them.
 //
-// Be sure to run the Calculate method on the results!
+// This method will call "Calculate" on th invoice automatically both before and after
+// to ensure that the data matches.
 //
-// To do this we need to figure out the "accuracy" or precision to use when removing the included
-// taxes so that we can avoid rounding errors. By default we add a single decimal place to all
-// numbers, but in the case of line items we take the length (Log 10) of the *quantity*.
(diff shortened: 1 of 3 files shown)


bill/invoice.go
```diff
@@ -209,42 +209,60 @@ func (inv *Invoice) Calculate() error {
...
+			return nil, err
+		}
 
-	return &i2
+		if inv.Totals.Total.String() == i2.Totals.Total.String() &&
+			inv.Totals.Tax.String() == i2.Totals.Tax.String() {
+			return &i2, nil
+		}
+	}
+	return nil, errors.New("insufficient precision, unable to remove included taxes")
```
cavalle, line 265: Just for reference: this error will still happen if the quantity is large (and complex) enough because the maximum precision we're allowing in the loop above 6. But there must be a limit and 6 is probably large enough (I guess it will be very unlikely to find quantities with more than 6 digits)

cavalle (approved):
LGTM. It's really a complex issue. I've investigated on my side and it doesn't seem posible to know in advance how many decimals you require to get back the original totals as that depends on the specific values involved, apparently. So, trial and error seems our best bet.

I also like from this PR that now we will get an earlier failure if the totals after removing taxes don't match those before removing them. This, as I discuss in my review comment, can still happen if the quantity is large (and "complex") enough.

bill/invoice.go
samlown: And thats 6 in addition to the base amount of 2, so 8 0s in total, e.g. 0.00000001 ...  it would indeed have to be a very extreme situation to hit that one!

PR #179, opened 2023-07-28
MX Discounts

cavalle:
* Validates that invoice-level discounts aren't provided (since they aren't allowed by the SAT)
* Adds a line-level discount to the invoice example
* Related to: https://github.com/invopop/gobl.cfdi/pull/11


samlown (approved):
👍  Are "charges" allowed?

cavalle:
> Are "charges" allowed?

Probably not. I'll open a subsequent PR with all the main GOBL features not allowed in MX.

PR #190, opened 2023-08-28
Fix tax calculation

cavalle:


tax/totals.go
```diff
@@ -232,6 +232,11 @@ func (rt *RateTotal) percentagesMatch(c *Combo) bool {
 	} else if rt.Surcharge != nil {
 		return false
 	}
+
+	if rt.Percent == nil {
+		return c.Percent == nil
```
samlown, line 237: Wouldn't this just be:
```suggestion
		return true
```

samlown (approved):

samlown:
I've merge this manually now! Thanks!

PR #191, opened 2023-08-29
Completing FR VAT rates, plus sources

samlown:
* France now has complete VAT rates.
* Added a `Source` model to the tax category so that we can start adding more structured data, such as official sources.


cavalle (approved):
LGTM!

PR #193, opened 2023-09-04
Complements

samlown:
* Adding `complements` property to bill.Invoice which uses an array of `schema.Object`
* `Header` now lives in the `head` package alongside the stamps.
* `gobl.Document` has been moved to the `schema` package and renamed to `schema.Object` with reduced complexity to avoid any dependencies on other GOBL structures.


samlown:
> I'm not 100% sure about the need for the `base` package, but I have no strong objection either.

Its required because Go doesn't support cyclic dependencies. Also worth noting that `base` would not appear in the JSON Schemas, they'd all still be inside the main GOBL namespace... which of course would break other autogenerated implementations in other languages that don't have support for cyclic dependencies.

But I wasn't convinced.

So I spent some time refactoring again. In the new approach there is no new `base` package, instead we have a new `head` package which contains the `Header` and `Stamp` objects. `Document` was moved into the `schema` package. Some of the digest stuff was moved up into the `Envelope`, along with the main document tests.

What do you think?

PR #194, opened 2023-09-04
Date Time support for cal package

samlown:
* Now with basic Date-Time support

### DIFF
--- a/build/schemas/cal/date-time.json
+++ b/build/schemas/cal/date-time.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://gobl.org/draft-0/cal/date-time",
+  "$ref": "#/$defs/DateTime",
+  "$defs": {
+    "DateTime": {
+      "type": "string",
+      "format": "date-time",
+      "title": "Date Time",
+      "description": "Civil date time in simplified ISO format with no time zone\ninformation, for example: 2021-05-26T13:45:00\n"
+    }
+  },
+  "$comment": "Generated with GOBL v0.55.0"
+}
--- a/build/schemas/cal/date-time.json
+++ b/build/schemas/cal/date-time.json
@@ -7,7 +7,7 @@
       "type": "string",
       "format": "date-time",
       "title": "Date Time",
-      "description": "Civil date time in simplified ISO format with no time zone\ninformation, for example: 2021-05-26T13:45:00\n"
+      "description": "Civil date time in simplified ISO format with no time zone\ninformation, for example: 2021-05-26T13:45:00"
     }
   },
   "$comment": "Generated with GOBL v0.55.0"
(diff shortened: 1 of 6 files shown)


cal/date_time.go
```diff
@@ -0,0 +1,186 @@
...
+				Hour:   hour,
+				Minute: minute,
+				Second: second,
+			},
+		},
+	}
+}
+
+// Now produces a new date time instance for the current UTC time.
+func Now() DateTime {
```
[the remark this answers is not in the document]
cavalle: I think both `cal.Timestamp.Now()` and `cal.DateTime.Now()` would make sense if that happens.

build/schemas/cal/date-time.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://gobl.org/draft-0/cal/date-time",
+  "$ref": "#/$defs/DateTime",
+  "$defs": {
+    "DateTime": {
+      "type": "string",
+      "format": "date-time",
+      "title": "Date Time",
+      "description": "Civil date time in simplified ISO format with no time zone\ninformation, for example: 2021-05-26T13:45:00\n"
```
cavalle, line 10: Do we want those line breaks in that description?

cavalle (approved):
LGTM! (just left a couple of minor comments)

cal/date_time.go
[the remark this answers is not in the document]
samlown: You can't do that in Go :-) But `cal.DateTimeNow()` is possible. I've renamed now to `ThisSecond()` which I find quite amusing and of course is aligned with Date's `Today()`.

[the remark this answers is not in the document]
cavalle: oh, of course! I wasn't thinking clearly about this method. I ❤️ `cal.ThisSecond()`

build/schemas/cal/date-time.json
samlown: I fixed this too. Any empty lines at the end of the here.Doc are deleted.

cavalle: I also meant the line breaks in the middle. Those line breaks are there originally to force text wrapping in a monospaced presentation (i.e. a code editor), but this is the JSON schema and that description could be used, for example, to automatically generate HTML documentation and we won't want a `<br>` or a `<p>` to be added at those places.

samlown: There isn't a great solution for that... Markdown will of course ignore single `\n` and they are not meaningful inside HTML, so I think they're safe to ignore.

cavalle: It doesn't feel 100% safe to assume those `\n` will always be ignored. Arguably, if those line breaks are expected to be ignored, then they shouldn't be there in the first place. We could write our own helper to remove new lines unless there are more than one consecutively.

samlown: While not the case yet, imagine if we had code samples inside the `here.Doc()`. You for sure don't want to remove the `\n` and I'm not sure what rules could be used to do that.

cavalle: ok then

PR #199, opened 2023-09-13
Refactoring rounding and calculations

samlown:
* Significant change to the way taxes and totals are calculated.
* Refined rounding so that precision is always maintained in all calculations, but the output is rounded to that of the currency.
* Incoporated `calculator` into `bill.Tax`
* Now possible to override the default calculation methods, especially useful for InvoiceXpress which uses line tax calculation and post-rounding on totals.
* Fixed bug whereby small negative cent amounts (less than `1.0`) where not being shown correctly
* Previous trail-and-error processing of removing included taxes from prices has now been removed.
* Incoportated a `rounding` property to the invoice totals (as per EN16931) which is used to indicate an amount to round the total by, this is used when removing taxes to ensure the final amount matches.


bill/invoice_test.go
```diff
@@ -516,19 +547,27 @@ func TestRemoveIncludedTaxDeep(t *testing.T) {
...
+	t.Logf("TOTALS: %v", string(data))
+	data, _ = json.Marshal(i2.Lines)
+	t.Logf("Lines: %v", string(data))
+	data, _ = json.Marshal(i2.Totals)
+	t.Logf("TOTALS: %v", string(data))
+
+	assert.Equal(t, "17830.20", i2.Totals.Total.String())
+	// assert.Equal(t, i.Totals.Total.String(), i2.Totals.Total.String())
+	assert.Equal(t, "17830.20", i2.Totals.Total.String())
+	assert.Equal(t, "-0.02", i2.Totals.Rounding.String())
```
samlown, line 570: This is a bit of a hack, but it ensures rounding errors after removing taxes are more or less handled.

PR #201, opened 2023-09-19
Removing total calculator option, increasing accuracy in line calculations

samlown:
* Simplifying the #199 commit by removing the line-by-line total calculation method which after some experimentation, appeared to make no difference to the results whatsoever.
* Sneaky change: Colombia now has support for simplified invoice tag. A few others support proforma.


cavalle (approved):
LGTM! Removing the total calculator option avoids a lot of complexity, I just hope we don't need to bring it back 🤞

As for the accuracy increase, changes look good and I tested they're effective. I'll just leave a thought here for your future consideration. The code is getting an increasing number of "rescaling" and "matching precision" invocations that make it harder and harder to understand. In my opinion, most of those rescales could be avoided if the Amount object would automatically rescale the results of operations to ensure no accuracy is lost. We could keep maximum accuracy for intermediate calculations by default, and do explicit rescaling in those cases where we actually want to round to a specific scale (like in totals).

I think I already suggested this in the past, and I still believe we could implement it with integers (i.e. no floating point) –at least for sums, diffs, and multiplications, which are the most common operations. If, at any point you think this is a path worth exploring, I'll be more than happy to discuss the approach.

samlown:
Yeah, I agree on the complexity, I'm also not convinced on the naming of the rescale and match precision methods.

We did try the auto-scaling in a previous version that would match the precision to whichever number is highest. This caused problems in the output which of course with the new rounding would not be a problem. I'll see if there is a quick win.

Regarding the integer vs float math, I think the current solution is okay. The only difference at the moment with floats is that multiplications and divisions will be rounded correctly in single operations whereas with integers numbers would always be rounded down, which could be a problem. The problem with floats is when you start adding them together.

PR #203, opened 2023-09-20
Correction Option Schema support

samlown:
* Opens up a new method `CorrectionOptionsSchema` that will generate a schema that can be used to dynamically load the available correction options for a given document.
* Initially only supports invoices.


cavalle (approved):
LGTM!

PR #207, opened 2023-09-28
Refactoring zones so that data is lazy loaded when needed

samlown:
* This fixes the bug with the WASM binary which couldn't cope with a large amount of raw data.
* Makes zone data loading more efficient.
* Unfortunately implies that the Region JSON data no longer contains the zones directly, they have to be loaded manually.


cavalle (approved):
Changes LGTM! If they work and fix the WASM issue, I wouldn't delay much releasing them.

Just a non-essential consideration: do we strictly need (to ensure WASM won't break) to also remove the zones from the `build/regimes` output files? If we don't, I think it will still be useful to have all the regime metadata in those files. They can be useful in the future to generate documentation, clients in other languages…

samlown:
@cavalle the regime zone data is something I was thinking about... the issue there is that both data sources would be included in the wasm with a significant file size increase (Colombia comes in at 256kb). In theory the raw data can still be extracted from each regions individual data files, but its not as good. I'm not sure what to do yet for this.

PR #209, opened 2023-09-29
Centralising data in a single repository

samlown:
* Moving all regime data into a single `/data/regimes` path.
* Schema JSON now also moved to `/data/schemas`.


cavalle (approved):
Replicating my feedback on Slack: "having everything under the same roof is an improvement, but I agree the solution doesn't feel like a well-rounded one. I cannot help but think that, as convenient as it is, maybe these potentially large catalogs of data shouldn't be part of the GOBL binary…"

PR #211, opened 2023-10-03
Food vouchers complement in MX

cavalle:
* Adds Food Vouchers complement to MX regime used to generate the CFDI's "ValesDeDespensa" complement.
* Related to the Fuel Account Balance complement (https://github.com/invopop/gobl/pull/210)
* The GOBL complements maps one by one to every node and field of the CFDI complement.
* Implements validations according to the constrains of the the ValesDeDespensa spec.
* Normalises all the amounts to the precision (2 decimals) specified in the ValesDeDespensa spec.
* Calculates the grand total.

### DIFF
--- a/regimes/mx/food_vouchers_complement.go
+++ b/regimes/mx/food_vouchers_complement.go
@@ -80,52 +80,37 @@ func (fvc *FoodVouchersComplement) Validate() error {
 			validation.Length(0, 20),
 		),
 		validation.Field(&fvc.Total, validation.Required),
-		validation.Field(&fvc.Lines,
-			validation.Required,
-			validation.Each(validation.By(validateFoodVouchersLine)),
-		),
+		validation.Field(&fvc.Lines, validation.Required),
 	)
 }
 
-func validateFoodVouchersLine(value interface{}) error {
-	line := value.(*FoodVouchersLine)
-	if line == nil {
-		return nil
-	}
-
-	return validation.ValidateStruct(line,
-		validation.Field(&line.EWalletID,
+func (fvl *FoodVouchersLine) Validate() error {
+	return validation.ValidateStruct(fvl,
+		validation.Field(&fvl.EWalletID,
 			validation.Required,
 			validation.Length(0, 20),
 		),
-		validation.Field(&line.IssueDateTime, cal.DateTimeNotZero()),
-		validation.Field(&line.Employee,
-			validation.Required,
-			validation.By(validateFoodVouchersEmployee)),
-		validation.Field(&line.Amount, validation.Required),
+		validation.Field(&fvl.IssueDateTime, cal.DateTimeNotZero()),
+		validation.Field(&fvl.Employee, validation.Required),
+		validation.Field(&fvl.Amount, validation.Required),
 	)
 }
 
-func validateFoodVouchersEmployee(value interface{}) error {
-	employee := value.(*FoodVouchersEmployee)
-	if employee == nil {
-		return nil
-	}
-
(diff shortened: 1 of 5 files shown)


regimes/mx/food_vouchers_complement.go
```diff
@@ -0,0 +1,122 @@
...
+	// Date and time of the food voucher's issue (maps to `Fecha`).
+	IssueDateTime cal.DateTime `json:"issue_date_time" jsonschema:"title=Issue Date and Time"`
+	// Employee's tax identity code (maps to `rfc`).
+	EmployeeTaxCode cbc.Code `json:"employee_tax_code" jsonschema:"title=Employee's Tax Identity Code"`
+	// Employee's CURP ("Clave Única de Registro de Población", maps to `curp`).
+	EmployeeCURP cbc.Code `json:"employee_curp" jsonschema:"title=Employee's CURP"`
+	// Employee's name (maps to `nombre`).
+	EmployeeName string `json:"employee_name" jsonschema:"title=Employee's Name"`
+	// Employee's Social Security Number (maps to `numSeguridadSocial`).
+	EmployeeSocialSecurity cbc.Code `json:"employee_social_security,omitempty" jsonschema:"title=Employee's Social Security Number"`
```
samlown, line 60: Whenever I see namespaces like this I think they could be inside their own object. Would a `FoodVoucherEmployee` object make sense?

samlown: I think so... we're renaming everything anyway from Spanish to English, so I do think its reasonable.

cavalle: Cool! Based on [previous feedback](https://github.com/invopop/gobl/pull/192#discussion_r1315048006), I was following the general rule of "stick[ing] as closely as possible to the original source structure, with translated fields". But, I agree that this is a minor deviation from that general rule that makes things nicer for the user. Addressed in 90eaae82147fe06b95f7322ecfb563971bf50cb5.

regimes/mx/food_vouchers_complement.go
```diff
@@ -0,0 +1,145 @@
...
+		),
+		validation.Field(&line.IssueDateTime, cal.DateTimeNotZero()),
+		validation.Field(&line.Employee,
+			validation.Required,
+			validation.By(validateFoodVouchersEmployee)),
+		validation.Field(&line.Amount, validation.Required),
+	)
+}
+
+func validateFoodVouchersEmployee(value interface{}) error {
```
samlown, line 109: BTW, you don't need to manually validate here. You just define the `func (fve *FoodVoucersEmployee) Validate() error` method and ensure the preceding structure includes the field definition:

```
    validation.Field(&line.Employee),
```

cavalle: Nice one! Addressed in cf73168

regimes/mx/food_vouchers_complement.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Validate checks the FoodVouchersLine data is valid.
+func (fvl *FoodVouchersLine) Validate() error {
+	return validation.ValidateStruct(fvl,
+		validation.Field(&fvl.EWalletID,
+			validation.Required,
+			validation.Length(0, 20),
+		),
+		validation.Field(&fvl.IssueDateTime, cal.DateTimeNotZero()),
+		validation.Field(&fvl.Employee, validation.Required),
```
samlown, line 95: This looks great now 👍

samlown (approved):
Looks great! 👍

PR #216, opened 2023-11-17
Retained and other MX taxes

cavalle:
* This PR completes the tax categories allowed in Mexico with: Retained VAT, ISR (only retained) and IEPS (both regular and retained versions)
* Related to https://github.com/invopop/gobl.cfdi/pull/17


samlown (approved):
Nice.

PR #217, opened 2023-11-20
Adding it inbox key definitions

samlown:
* Support for inbox key definitions.
* No consequences at the moment other than better aesthetics and attempt to standardise use of specific codes.

### DIFF
--- a/data/regimes/it.json
+++ b/data/regimes/it.json
@@ -918,6 +918,22 @@
       }
     }
   ],
+  "inbox_keys": [
+    {
+      "key": "it-sdi-code",
+      "name": {
+        "en": "SDI Desination Code",
+        "it": "Codice Destinatario SDI"
+      }
+    },
+    {
+      "key": "it-sdi-pec",
+      "name": {
+        "en": "SDI PEC Destination",
+        "it": "PEC Destinatario SDI"
+      }
+    }
+  ],
   "scenarios": [
     {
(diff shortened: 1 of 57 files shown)


data/regimes/it.json
```diff
@@ -918,6 +918,22 @@
       }
     }
   ],
+  "inbox_keys": [
+    {
+      "key": "it-sdi-code",
+      "name": {
+        "en": "SDI Desination Code",
```
cavalle, line 925:
```suggestion
        "en": "SDI Destination Code",
```

regimes/it/examples/hotel.json
```diff
@@ -39,7 +39,7 @@
     "name": "Mela S.r.l.",
     "inboxes": [
       {
-        "key": "codice-destinatario",
+        "key": "it-sdi-code",
```
cavalle, line 42: Mentioning just in case: Is it safe to assume this won't break any existing implementation (e.g. Amenitiz) and that we don't need a migration?

cavalle (approved):
Looking good! Just left a couple of comments

regimes/it/examples/hotel.json
samlown: So far, no one has used this in production, so I think we're good. Previous support was also a bit hacky without a specific constant to use. It'll be easy to add to the gobl.fatturapa tool though if anyone needs it.

PR #218, opened 2023-11-20
Defining common invoice tags, simplified invoice validation moved up

samlown:
* Simplified invoice validation is now universal.
* Defined a standard set of invoice tags that can be used potentially in any invoice, with the possibility for tax regimes to add to their own local definition.
* Default tags are:
  - `simplified` for simplified invoices
  - `reverse-charge` for when the recipient is responsible for tax
  - `self-billed` for invoices created by the customer on behalf of the supplier.
  - `customer-rates` indicates that the tax rates of the customer's region should be applied instead of the supplier, such as for digital goods sold to B2C customers.


juanmoli9:
@samlown Food for thought
> `customer-rates` indicates that the tax rates of the customer's region should be applied instead of the supplier, such as for digital goods sold to B2C customers. 

Shouldn't the default tax rates be those of the customer? Most countries adhere to the destination principle, which is the official recommendation by OECD. 

> Guideline 3.1: For consumption tax purposes internationally traded services and intangibles should be taxed according to the rules of the jurisdiction of consumption.

Source: [International VAT/GST Guidelines - Chapter 1.C](https://www.oecd-ilibrary.org/docserver/9789264271401-en.pdf?expires=1700653325&id=id&accname=guest&checksum=599ADAEDB36B20EE6506EBA22C39670F)

samlown:
@juanmoli9 I see your point... this is a semantic issue and no one is currently using this, so the names can be changed.

What could be the alternative? Ideally we want something explicit as its hard to guess if the final customer is a business or physical person based on the data. Perhaps a simple tag like "`b2c`"?

build/schemas/regimes/mx/food-vouchers.json
cavalle, line 1: @samlown did you remove these two JSONs intentionally? Without them, we won't have the docs at https://docs.gobl.org/draft-0/regimes/mx/food_vouchers

samlown: Yes, because the output path has changed. Not quite sure why there were in the old "build" directory instead of "data".

cavalle: Ok, right! Thanks

PR #219, opened 2023-11-21
Addenda Mabe in MX regime

cavalle:
* Adds extensions and validations necessary to generate "Addenda Mabe" in the MX regime
* Addenda Mabe generation implemented in invopop/gobl.cfdi#18

### DIFF
--- a/bill/invoice.go
+++ b/bill/invoice.go
@@ -95,6 +95,9 @@ type Invoice struct {
 	// Additional complementary objects that add relevant information to the invoice.
 	Complements []*schema.Object `json:"complements,omitempty" jsonschema:"title=Complements"`
 
+	// Extension code map for any additional regime specific codes that may be required.
+	Ext cbc.CodeMap `json:"ext,omitempty" jsonschema:"title=Ext"`
+
 	// Additional semi-structured data that doesn't fit into the body of the invoice.
 	Meta cbc.Meta `json:"meta,omitempty" jsonschema:"title=Meta"`
 }
@@ -144,6 +147,7 @@ func (inv *Invoice) ValidateWithContext(ctx context.Context) error {
 		validation.Field(&inv.Totals, validation.Required),
 
 		validation.Field(&inv.Notes),
+		validation.Field(&inv.Ext, tax.InRegimeExtensions),
 		validation.Field(&inv.Meta),
 
 		validation.Field(&inv.Complements, validation.Each()),
--- a/regimes/mx/extensions.go
+++ b/regimes/mx/extensions.go
@@ -1,6 +1,7 @@
 package mx
 
 import (
+	"github.com/invopop/gobl/bill"
 	"github.com/invopop/gobl/i18n"
 	"github.com/invopop/gobl/tax"
 )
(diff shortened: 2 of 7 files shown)


bill/invoice.go
```diff
@@ -95,6 +95,9 @@ type Invoice struct {
 	// Additional complementary objects that add relevant information to the invoice.
 	Complements []*schema.Object `json:"complements,omitempty" jsonschema:"title=Complements"`
 
+	// Extension code map for any additional regime specific codes that may be required.
+	Ext cbc.CodeMap `json:"ext,omitempty" jsonschema:"title=Ext"`
```
cavalle, line 99: I couldn't find a good place to store Mabe's `reference1` and `reference2` fields so I added this Ext codeMap at invoice level. The content of `reference1` is a bit ambiguous, and changes depending on the case:

image

In the case of `reference2` this is what the doc says: "Referencia adicional para uso futuro o bien "NA" o "0"".

Suggestions on a better place to put this data are welcome!

cavalle: See the `addenda-mabe.yaml` below to see it in context

regimes/mx/extensions.go
```diff
@@ -333,4 +343,66 @@ var extensionKeys = []*tax.KeyDefinition{
...
+		Desc: i18n.String{
+			i18n.EN: "Reference 2 used by Mabe.",
+			i18n.ES: "Referencia 2 utilizada por Mabe.", //nolint:misspell
+		},
+	},
+}
+
+// IsMabeSupplier returns true if the invoice is from a Mabe supplier.
+func IsMabeSupplier(inv *bill.Invoice) bool {
+	return inv != nil && inv.Supplier != nil && inv.Supplier.Ext.Has(ExtKeyMabeProviderCode)
```
cavalle, line 407: This function encapsulates the criteria to decide whether a MX GOBL invoice will produce a Mabe Addendum: if the supplier provides it's Mabe Supplier code, GOBL will validate that all the data required to generate the Mabe Addendum is present and the CFDI library will generate and attach the addendum to the CFDI.

regimes/mx/invoice_validator.go
```diff
@@ -106,6 +121,20 @@ func (v *invoiceValidator) validLine(value interface{}) error {
 			validation.Required,
 			validation.Skip, // Prevents each tax's `ValidateWithContext` function from being called again.
 		),
+		validation.Field(&line.Item, validation.By(v.validItem)),
+	)
+}
+
+func (v *invoiceValidator) validItem(value interface{}) error {
```
cavalle, line 128: This one was a bit tricky. We were already validating the Item in the MX regime separately, however I cannot add these validations there because I need to call to the `IsMabeSupplier` function with the invoice as a parameter. In the Item standalone validation, naturally, we don't have access to invoice. That's why I implemented this particular validation as part of the invoice.

Suggestion on how to do this differently are more than welcome.

cavalle:
We decided to keep MX addendas out of GOBL. It won't be easy to scale to potentially hundreds or thousand of potential addendas.

PR #223, opened 2023-11-30
Unit updates and compatibility with UN/ECE codes

samlown:
* Adds a few more "frequently used" unit keys for Mexico
* Brings support for being able to use either a unit key *or* UN/ECE unit code in the unit value (not entirely sure yet if this is genius or stupid.)

### DIFF
--- a/data/schemas/org/unit.json
+++ b/data/schemas/org/unit.json
@@ -4,219 +4,303 @@
   "$ref": "#/$defs/Unit",
   "$defs": {
     "Unit": {
-      "anyOf": [
+      "oneOf": [
+        {
+          "const": "mg",
+          "title": "Milligrams"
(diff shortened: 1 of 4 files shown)


samlown:
@cavalle I hear you, but I'm not convinced on the alternate paths at the moment. Also bare in mind that the previous version allowed *any* key in the unit field. You'll see in the commit there that some of the tests in Mexico were using an invalid unit, which wasn't great either.

While it feels a bit unusual, the JSON Schema is sound. Its just a string for which you can use a constant, or a code. The unit isn't used for decision making, which lessons the requirements, and I think its useful to be able to fall back on standard codes if the customer requires it.

I'll commit for the time being so we can fix some of the issues in Mexico.

PR #228, opened 2023-12-31
Maintain precision in category with taxes included

samlown:
* Fixes rounding issue calculating totals when prices include tax.


cavalle (approved):
Looks good to me! I'll rollback the 1 cent tolerance in the IX provider, when this is merged

PR #231, opened 2024-01-16
Change and document payment method and means in MX

cavalle:
* Documents how payment method (`MethodPago`) is inferred from the presence or absence of advances
* Expects the payment means keys to be given in the advances (instead of in the instructions) when an invoice is prepaid to map to a `FormaPago` value. 
* Doesn't expect a payment means key in non-prepaid invoices. The `FormaPago` in that case must always be set `99` (Por definir)
* Allows the payment means key to be given in the instructions temporarily until the users transition to use the advances

### DIFF
--- a/regimes/mx/README.md
+++ b/regimes/mx/README.md
@@ -119,9 +119,9 @@ The following GOBL maps to the `G03` (Gastos en general) value of the `UsoCFDI` 
 
 The CFDI’s `MetodoPago` field specifies whether the invoice has been fully paid at the moment of issuing the invoice (`PUE` - Pago en una sola exhibición) or whether it will be paid in one or several instalments after that (`PPD` – Pago en parcialidades o diferido).
 
-In GOBL, the presence or absence of payment advances covering for the invoice’s total payable amount will denote whether the `MetodoPago` will be set to `PUE` or `PPD`.
+In GOBL, the presence or absence of payment advances covering the invoice’s total payable amount will determine whether the `MetodoPago` will be set to `PUE` or `PPD`.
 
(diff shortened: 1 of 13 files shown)


regimes/mx/README.md
```diff
@@ -115,11 +115,94 @@ The following GOBL maps to the `G03` (Gastos en general) value of the `UsoCFDI`
 }
 ```
 
+### `MetodoPago` – Payment Method
+
+The CFDI’s `MetodoPago` field specifies whether the invoice has been fully paid at the moment of issuing the invoice (`PUE` - Pago en una sola exhibición) or whether it will be paid in one or several instalments after that (`PPD` – Pago en parcialidades o diferido).
+
+In GOBL, the presence or absence of payment advances covering for the invoice’s total payable amount will denote whether the `MetodoPago` will be set to `PUE` or `PPD`.
+
+Please note that if you don't include any advances in your GOBL invoice, it will be assumed that the payment of the invoice is outstanding (`MetodoPago = PPD`). This implies that the invoice supplier will have to issue “CFDI de Complemento de Pago” documents at a later stage when receiving the invoice payments.
```
samlown, line 124:
```suggestion
Please note that if you don't include any advances in your GOBL invoice, it will be assumed that the payment of the invoice is outstanding (`MetodoPago = PPD`). This implies that the invoice supplier will have to issue “CFDI de Complemento de Pago” (remittance advice) documents at a later stage when receiving the invoice payments.
```

samlown (approved):
Looks good. A few possible copy suggestions. I like the approach with the advance key checks, makes sense. 👍

cavalle:
Thanks a lot @samlown! Copy suggestions are always very much appreciated!

PR #232, opened 2024-01-17
Support for signature and header verification

samlown:
* Adds a "Verify" method to the envelope that makes it easier to check an optional list of keys against the signatures used to sign the document.
* `head.Header` now has a "Contains" method that facilitates testing if the current header includes all the fields from the provided header.

### DIFF
--- a/dsig/signature.go
+++ b/dsig/signature.go
@@ -133,28 +133,42 @@ func (s *Signature) String() string {
 	return d
 }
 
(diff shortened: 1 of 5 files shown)


cavalle (commented):
Handy methods! Looking great

samlown:
The Header "contains" method is a bit bespoke, I was hoping for something more generic, but this felt like the best way for the time being to handle comparisons. I'm surprised we didn't have have this earlier though 😆 🤦

PR #234, opened 2024-01-22
Establish consistent usage of 'desc' in public models

samlown:
* Removing usage of `"desc"` property name in favour of `"description"` which is used more widely throughout GOBL and is consistent with JSON Schema definitions.

### DIFF
--- a/org/item.go
+++ b/org/item.go
@@ -33,8 +33,8 @@ type Item struct {
 	Name string `json:"name"`
 	// List of additional codes, IDs, or SKUs which can be used to identify the item. They should be agreed upon between supplier and customer.
 	Identities []*Identity `json:"identities,omitempty" jsonschema:"title=Identities"`
-	// Detailed description
-	Description string `json:"desc,omitempty"`
+	// Detailed description of the item.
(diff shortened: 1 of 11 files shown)


org/item.go
cavalle, line 1: No custom `UnmarshalJSON` for this struct as well?

cavalle (commented):
Changes LGTM (just left one comment). Regarding impact on modules/providers, if I'm not mistaken all our modules use either Go's `gobl` or `gobl.ruby` to access GOBL documents. That means that we would need to:
1. Prepare a change in `gobl.ruby` updating to the latest GOBL schema and adding backwards compatibility with the `desc` name.
2. Update all the modules (or, at least, all that access to any description field, e.g. CO, MX…) to use the latest version of `gobl` or `gobl.ruby` to ensure they be able to process documents with both `desc` and `description` fields.

cavalle (approved):

org/item.go
samlown: I think my though process here was that we've not seen anybody use this field, so its not worth the extra code... have you seen any exceptions?

PR #236, opened 2024-01-26
Update Tax ID validations in CO

cavalle:
* Ensure all the Tax ID validations previously done in the provider are now in GOBL (related to https://github.com/invopop/plemsi/pull/51)
* Validate the Tax ID properly for foreign customers

### DIFF
--- a/regimes/co/invoices.go
+++ b/regimes/co/invoices.go
@@ -48,21 +48,17 @@ func (v *invoiceValidator) validate() error {
 
 func (v *invoiceValidator) validParty(value interface{}) error {
 	obj, _ := value.(*org.Party)
-	if obj == nil || obj.TaxID == nil {
+	if obj == nil {
 		return nil
 	}
 	return validation.ValidateStruct(obj,
 		validation.Field(&obj.TaxID,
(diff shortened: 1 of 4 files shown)


samlown (commented):
I'm not sure about this one... doesn't this imply that GOBL user's must know that the code for end-customers should be `2222...`? Also, what happens in the case of simplified invoices when the customer object is missing?

cavalle:
> I'm not sure about this one... doesn't this imply that GOBL user's must know that the code for end-customers should be 2222...? Also, what happens in the case of simplified invoices when the customer object is missing?

@samlown For the final customer case, the Plemsi provider supports both providing an explicit customer with the `222…` code (and without a zone) or excluding the customer and using the `simplified` tax tag. This last option was added later and we have one client that built its integration before it that is still using the `222…` code explicitly. We can totally remove support for the explicit case: we'd just need to coordinate with the client to migrate (or implementing some internal normalisation). 

I don't have a strong opinion, I think it's fine to keep supporting both, although I also see the appeal of keeping things simple and consistent supporting just the `simplified` case. Let me know what you think.

PR #237, opened 2024-01-30
Basic templating support example

samlown:
* Setting up an example for using templates to be able to merge data in other formats and generate documents.


cavalle (approved):
LGTM as a start to enhance as we need!

PR #241, opened 2024-02-06
VAT always required in Italy

samlown:
* Quick test to ensure that invoice lines in Italy always have VAT defined, even if exempt.


cavalle (approved):
LGTM!

PR #244, opened 2024-02-08
Add TBAI preceding validations

cavalle:
* Adds validations for the preceding element of TBAI corrective invoices


samlown (approved):
Looks good! Unlike you though @cavalle to submit without tests :-) Related to that, we do actually need to increase test coverage so that we can submit GOBL to different lists of open source projects, we can add those later.

PR #245, opened 2024-02-08
Introducing a recommended flag to json Schema

samlown:
* Adds a "recommended" array alongside "required" that lists properties that should be presented in User Interfaces, but are not required.


cavalle (commented):
LGTM!

PR #247, opened 2024-02-14
Validate presence of supplier address in IT regime

cavalle:
* Adds validation for the presence of a supplier address that will later be used to populate the `condense_prestatore.sede` field in FaturraPA.
* Prevents a crash happening when an empty array of addresses were given in a party

This was causing these error in production:

image

Reference from the FatturaPA docs:

image

### DIFF
--- a/regimes/it/invoice_validator.go
+++ b/regimes/it/invoice_validator.go
@@ -63,7 +63,8 @@ func (v *invoiceValidator) supplier(value interface{}) error {
 			tax.IdentityTypeIn(TaxIdentityTypeBusiness, TaxIdentityTypeGovernment),
 		),
 		validation.Field(&supplier.Addresses,
-			validation.By(validateAddress),
+			validation.Required,
+			validation.Each(validation.By(validateAddress)),
 		),
 		validation.Field(&supplier.Registration,
 			validation.By(validateRegistration),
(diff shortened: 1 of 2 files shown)


regimes/it/invoice_validator.go
```diff
@@ -109,15 +110,14 @@ func isItalianParty(party *org.Party) bool {
...
 		return nil
 	}
 	// Post code and street in addition to the locality are required in Italian invoices.
-	address := v[0]
-	return validation.ValidateStruct(address,
-		validation.Field(&address.Street, validation.Required),
-		validation.Field(&address.Code, validation.Required),
+	return validation.ValidateStruct(v,
+		validation.Field(&v.Street, validation.Required),
+		validation.Field(&v.Code, validation.Required),
```
samlown, line 120: Technically the Code is only required for Italian addresses, it'll be replaced with `00000` for non-Italian addresses. Not sure if that is worth considering here.

samlown (approved):
LGTM. Just a comment on the Code (which was already implemented)

regimes/it/invoice_validator.go
cavalle: At the moment, this validation [applies only to Italian customers](https://github.com/invopop/gobl/blob/a07bb8222f5ea1280d9bbca1b1d2c3485544b0d6/regimes/it/invoice_validator.go#L95-L100). That check is not done for suppliers though. I'm guessing an assumption was made here that suppliers will always have an Italian address. Sounds like a fair assumption to me, but we can always move the `isItalianParty` condition to this very method to apply to both.

PR #250, opened 2024-02-27
Normalizing ExtMap, Stamps, and others to remove empty rows

samlown:
* ExtMap values are now normalized, removing any empty values.
* Stamps also normalized
* UUID in `org.Party` is now cleaning UUID.

### DIFF
--- a/bill/invoice.go
+++ b/bill/invoice.go
@@ -196,8 +196,9 @@ func (inv *Invoice) Empty() {
 	inv.Payment.ResetAdvances()
 }
 
-// Calculate performs all the calculations required for the invoice totals and taxes. If the original
-// invoice only includes partial calculations, this will figure out what's missing.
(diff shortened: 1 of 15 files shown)


cavalle (approved):
Looking great! Keeping GOBL files nice and tidy FTW :)

I still find surprising `ExtMap` is under `tax` and not under a more generic package such us `cbc`, but I believe we already had and settled that discussion.

samlown:
> I still find surprising `ExtMap` is under `tax` and not under a more generic package such us `cbc`, but I believe we already had and settled that discussion.

I think the main confusion there is with the values, I'm not happy with the `KeyOrCode` type. I'm tempted to convert it into a `string` without strict validation: most extensions have a fixed list of constants anyway.

I'm also tempted to rename into simply `Extensions` instead of `ExtMap`. That way the implications of `tax.Extensions` is way more obvious.

PR #251, opened 2024-02-27
Refactoring ExtMap to Extensions

samlown:
* Renaming `tax.ExtMap` to `tax.Extensions`
* Removing the `cbc.KeyOrCode` which was always a lame duck.
* Adding `tax.ExtValue` for internal use when handling extension values.
* Change implies that extensions can now potentially be any string, which is potentially useful for long values like descriptions or reasons.


cavalle (approved):
Nice!

PR #253, opened 2024-02-28
Add UUIDv3 creation functions

cavalle:
* Useful to create UUIDs from other external identifiers deterministically

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -25,6 +25,11 @@ func MakeV1() UUID {
 	return UUID{uuid.Must(uuid.NewUUID())}
 }
 
+// MakeV3 generates a new MD5 UUID using the provided namespace and data.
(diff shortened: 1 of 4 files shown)


samlown (commented):
I think I'd suggest adding some tests for these cases! :-) Since we're here, it might also make senes to support UUIDv5, and also add some comments about the differences. (v3 is faster, but less secure apparently)

cavalle:
> I think I'd suggest adding some tests for these cases! :-) Since we're here, it might also make sense to support UUIDv5, and also add some comments about the differences. (v3 is faster, but less secure apparently)

Thanks! I'll do all that

cavalle:
Just realised the comments aren't correct. A hash can never be "reversed", I'll fix it

cavalle:
Fixed in: b90dc64

PR #254, opened 2024-03-13
Add correction invoice fields for Polish regime

noplisu:
## Pull Request Summary
To handle credit notes (correction invoices) we need to add fields required by the Polish regime (KSeF). Those fields are:
- The KSeF number of the invoice being corrected if the corrected invoice was submitted to KSeF
- The correction type
- Reason for the correction


regimes/pl/invoices.go
```diff
@@ -27,7 +27,7 @@ func (v *invoiceValidator) validate() error {
 		// Rectification state determined by Preceding value.
 		validation.Field(&inv.Type, validation.In(
 			bill.InvoiceTypeStandard,
-			// bill.InvoiceTypeCorrective,
+			bill.InvoiceTypeCorrective,
```
samlown, line 30: So my initial impression here is that these in fact should be "Credit Notes" as opposed to Corrective invoices. This is a confusing subject as some countries, like Spain, don't recognise the "credit notes" as such, but instead define the concept of corrective invoices and a replacement strategy.

From the looks of things here and the example with negative values, I'd suggest using the `bill.InvoiceTypeCreditNote` type instead.

The difference in GOBL is as follows:
 * Corrective invoices *replace* the previous document.
 * Credit (or Debit) notes *extend* the previous document.

It may indeed be that Poland supports *both* types.

regimes/pl/extensions.go
```diff
@@ -64,4 +65,45 @@ var extensionKeys = []*tax.KeyDefinition{
 			},
 		},
 	},
+	{
+		Key: ExtKeyKSEFCorrection,
+		Name: i18n.String{
+			i18n.EN: "Zero VAT Extensions for KSeF",
+		},
```
samlown, line 72: So I'm confused with the naming here, I suspect it was most likely copied and pasted incorrectly :-) However, the descriptions below suggest that this extension code has more to do with "date" than configuration of the corrective document. I've not fully understood the meaning. Do you have any more context?

regimes/pl/pl.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/invopop/gobl/tax"
 )
 
+// KSeF official codes to include.
+const (
+	StampProviderKSeF cbc.Key = "ksef-id"
+	CorrectionType    cbc.Key = "cor-type"
```
samlown, line 17: I don't think this value is used anywhere.

regimes/pl/pl.go
```diff
@@ -39,6 +45,21 @@ func New() *tax.Regime {
...
+				Schema: bill.ShortSchemaInvoice,
+				Types: []cbc.Key{
+					bill.InvoiceTypeCorrective,
+				},
+				ReasonRequired: true,
+				Stamps: []cbc.Key{
+					StampProviderKSeF,
+				},
+				Extensions: []cbc.Key{
+					ExtKeyKSEFCorrection,
```
samlown, line 59: If this extension field is required, it might also need adding to invoice validation.

samlown (changes requested):
I think there are a few things here that need elaborating to make it a bit easier to understand.

regimes/pl/invoices.go
noplisu: It definitely extends the previous invoice. I'll change the type.

regimes/pl/extensions.go
noplisu: Yeah, I forgot to change this description.. This is a field In KSeF which informs the tax authority on when the correction should take legal effect, with values 1-3 meaning:
1 - Changes take effect in the date of the invoice. This is mostly used if you made a mistake in the original invoice.
2 - Changes take effect in the date of the correction. This is used for example when giving a discount or returning items.
3 - Changes take effect in a different date than any of the invoices or the date is different for different elements.

regimes/pl/pl.go
noplisu: I added validation for the extension field with the tests.

regimes/pl/examples/corrective-invoice-pl-pl.yaml
```diff
@@ -0,0 +1,56 @@
+$schema: "https://gobl.org/draft-0/bill/invoice"
+currency: "PLN"
+issue_date: "2023-12-21"
+series: "COR"
+code: "002"
+type: "corrective"
```
samlown, line 6: As far as I understand, this type of invoice should not be supported, it should be `credit-note`.

regimes/pl/examples/corrective-invoice-pl-pl.yaml
```diff
@@ -0,0 +1,56 @@
...
+      locality: "Madrid"
+      region: "Madrid"
+      code: "00-015"
+      country: "PL"
+
+lines:
+  - quantity: 20
+    item:
+      name: "Development services"
+      price: "-10.00"
```
samlown, line 52: Given that this should be a credit-note, the prices should be positive.

regimes/pl/extensions.go
```diff
@@ -9,6 +9,7 @@ import (
 const (
 	ExtKeyKSeFVATZero    = "pl-ksef-vat-zero"
 	ExtKeyKSeFVATSpecial = "pl-ksef-vat-special"
+	ExtKeyKSEFCorrection = "pl-ksef-correction"
```
samlown, line 12: Correction is too generic, I'd propose:
```suggestion
	ExtKeyKSEFEffectiveDate = "pl-ksef-effective-date"
```

regimes/pl/extensions.go
```diff
@@ -64,4 +65,46 @@ var extensionKeys = []*cbc.KeyDefinition{
 			},
 		},
 	},
+	{
+		Key: ExtKeyKSEFCorrection,
+		Name: i18n.String{
+			i18n.EN: "Correction type extension for KSeF credit notes",
+			i18n.PL: "Typ korekty dla faktur korygujących w KSeF",
```
samlown, line 72:
```suggestion
			i18n.EN: "Effective date code",
			i18n.PL: "Kod daty wejścia w życie",
```
?

regimes/pl/extensions.go
```diff
@@ -64,4 +65,46 @@ var extensionKeys = []*cbc.KeyDefinition{
...
+		},
+		Codes: []*cbc.CodeDefinition{
+			{
+				Code: "1",
+				Name: i18n.String{
+					i18n.EN: "Original",
+					i18n.PL: "Pierwotna",
+				},
+				Desc: i18n.String{
+					i18n.EN: "Correction has legal consequences in the date of the original invoice",
```
samlown, line 82:
```suggestion
					i18n.EN: "Effective as of the issue date of the original invoice.",
```

regimes/pl/extensions.go
```diff
@@ -64,4 +65,46 @@ var extensionKeys = []*cbc.KeyDefinition{
...
+				},
+			},
+			{
+				Code: "2",
+				Name: i18n.String{
+					i18n.EN: "Correction",
+					i18n.PL: "Korygująca",
+				},
+				Desc: i18n.String{
+					i18n.EN: "Correction has legal consequences in the date of the correction invoice",
```
samlown, line 93:
```suggestion
					i18n.EN: "Effective according to date of correction.",
```

samlown (commented):
A few bits to consider! looking good in general though.

regimes/pl/README.md
```diff
@@ -53,12 +53,12 @@ The following GOBL maps to the `1` (gotówka) value of the `TFormaPlatnosci` fie
...
-| KOR_ZAL | `corrective` | `partial`    | Corrective (advance)     |
-| KOR_ROZ | `corrective` | `settlement` | Corrective (settlement)  |
+| Code    | Type          | Tax Tags     | Description               |
+| ------- | ------------- | ------------ | ------------------------- |
+| VAT     | `standard`    |              | Regular invoice           |
+| UPR     | `standard`    | `simplified` | Simplified (no customer)  |
+| ZAL     | `standard`    | `partial`    | Advance invioce           |
+| ROZ     | `standard`    | `settlement` | Settlement invoice        |
+| KOR     | `credit-note` |              | Credit note (regular)     |
+| KOR_ZAL | `credit-note` | `partial`    | Credit note (advance)     |
```
samlown, line 63: It might be worth clarifying here if possible. Is this a "partial" credit note, or a credit note of a partial invoice?

regimes/pl/extensions.go
```diff
@@ -7,8 +7,9 @@ import (
 
 // Regime extension codes for local electronic formats.
 const (
-	ExtKeyKSeFVATZero    = "pl-ksef-vat-zero"
-	ExtKeyKSeFVATSpecial = "pl-ksef-vat-special"
+	ExtKeyKSeFVATZero       = "pl-ksef-vat-zero"
+	ExtKeyKSeFVATSpecial    = "pl-ksef-vat-special"
+	ExtKeyKSEFEffectiveDate = "pl-ksef-effective-date"
```
cavalle, line 12: Based on [Greg's this description](https://github.com/invopop/gobl/pull/254#discussion_r1528608583), I think this is a good name for the extension. The only (minor) drawback is that the name could suggest that the content will be an actual date. We could avoid that by making the name longer but I think it's preferable to keep it succinct. I wouldn't change it.

I spotted a typo in the constant name, though:
```suggestion
	ExtKeyKSeFEffectiveDate = "pl-ksef-effective-date"
```

cavalle (approved):
Changes LGTM!

regimes/pl/README.md
noplisu: Good question :D. This is a credit note of a partial invoice

samlown (approved):
👍

PR #264, opened 2024-04-03
Custom formatting for negative numbers

samlown:
* Fixes bugs when formatting negative numbers
* `num.Formatter` now allows for a `NegativeTemplate` property to be able to output `-10.20` as `(10.20)` for example.
* Added more tests around numbers in general.
* `num.Percentage` no longer extends `num.Amount`, instead it defines its own methods and provides an `Amount()` method that will respond with the factored percentage value, i.e. if the percentage is `16.0%`, the `Amount()` method will provide `16.0`.

### DIFF
--- a/num/amount.go
+++ b/num/amount.go
@@ -1,7 +1,6 @@
 package num
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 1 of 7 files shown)


cavalle (approved):
LGTM!

PR #267, opened 2024-04-12
Clean up tax categories

noplisu:
### Dependencies
This PR is a dependency for https://github.com/invopop/gobl.ksef/pull/14

## Pull Request Summary
To handle additional tax categories we need to clean up tax categories and extensions. We want to handle tax categories for:
```
standard                      - tax.RateStandard
reduced                       - tax.RateReduced
super reduced                 - tax.RateSuperReduced
taxi                          - tax.RateSpecial + ExtKeyKSeFVATSpecial{taxi}
inside Poland 0%              - tax.RateZero
inside EU 0%                  - tax.RateZero
outside EU 0%                 - tax.RateZero
tax exempt                    - tax.RateExempt
tax not applicable outside EU - TaxRateNotPursuant
tax not applicable inside EU  - TaxRateNotPursuant

Additional values to determine which case it is will be determined in the gobl.ksef package
```

## When do we use the zero tax rate
The regulations indicate situations in which a business entity may apply a 0% VAT rate for sales. It is essential for the seller to be an active VAT taxpayer. The 0% rate does not mean exemption from taxation. Sales made with a 0% VAT rate are still considered taxable. Despite no tax due, sales at a zero VAT rate should be reported in the VAT declaration. The 0% rate can be applied, among others, for the following activities:
- Intra-Community Supply of Goods (ICS),
- export of goods,
- services related to maritime and air transport.

## When do we use the tax exempt tax rate
The "ZW" VAT invoice is used by entrepreneurs who are not active VAT taxpayers, are exempt from taxation on goods and services on a subjective basis, or are active VAT taxpayers who provide VAT-exempt services. The "ZW" VAT invoice used by entrepreneurs for the sale of goods or services rendered does not allow for the deduction of VAT on purchases covered by "ZW". Activities exempt from tax include, for example:
- delivery of own agricultural products,
- educational services,
- social assistance services,
- financial and insurance intermediary services,
- services related to sports and physical education.

## When do we use the tax not applicable tax rate
The entrepreneur who includes the annotation "NP" in the field corresponding to the VAT rate on the invoice indicates that the sale to which the entry refers is not subject to taxation in Poland. The designation "NP" is an abbreviation for the phrase "not subject to." This situation occurs, for example, in the case of entrepreneurs who provide services related to export for foreign customers. Taxation will occur in the buyer's country. The "NP" annotation allows for the settlement of VAT on purchases covered by the entry.

### DIFF
--- a/regimes/pl/extensions.go
+++ b/regimes/pl/extensions.go
@@ -7,7 +7,6 @@ import (
 
 // Regime extension codes for local electronic formats.
 const (
-	ExtKeyKSeFVATZero       = "pl-ksef-vat-zero"
 	ExtKeyKSeFVATSpecial    = "pl-ksef-vat-special"
 	ExtKeyKSeFEffectiveDate = "pl-ksef-effective-date"
 )
@@ -30,38 +29,6 @@ var extensionKeys = []*cbc.KeyDefinition{
 					i18n.EN: "Special flat rate for taxi drivers.",
 					i18n.PL: "Specjalna stawka ryczałtu dla taksówkarzy.",
 				},
-			},
-		},
-	},
-	{
-		Key: ExtKeyKSeFVATZero,
-		Name: i18n.String{
-			i18n.EN: "Zero VAT Extensions for KSeF",
-		},
-		Keys: []*cbc.KeyDefinition{
-			{
-				Key: "wdt",
-				Name: i18n.String{
-					i18n.EN: "WDT",
-					i18n.PL: "WDT",
-				},
-				// TODO: description required
-			},
-			{
-				Key: "domestic",
-				Name: i18n.String{
-					i18n.EN: "Domestic",
-					i18n.PL: "Krajowy",
-				},
-				// TODO: description required
-			},
-			{
-				Key: "export",
-				Name: i18n.String{
-					i18n.EN: "Export",
-					i18n.PL: "Eksport",
-				},
-				// TODO: description required
 			},
 		},
 	},
--- a/data/regimes/pl.json
+++ b/data/regimes/pl.json
@@ -89,30 +89,42 @@
       ]
     },
     {
-      "key": "pl-ksef-vat-zero",
+      "key": "pl-ksef-vat-region",
       "name": {
-        "en": "Zero VAT Extensions for KSeF"
+        "en": "Region VAT Extensions for KSeF"
       },
-      "keys": [
+      "codes": [
         {
-          "key": "wdt",
+          "code": "EU",
           "name": {
-            "en": "WDT",
-            "pl": "WDT"
+            "en": "EU",
+            "pl": "EU"
+          },
+          "desc": {
+            "en": "Intra-EU Transactions",
+            "pl": "Transakcje wewnątrzwspólnotowe"
           }
         },
         {
-          "key": "domestic",
+          "code": "domestic",
           "name": {
             "en": "Domestic",
             "pl": "Krajowy"
+          },
+          "desc": {
+            "en": "Domestic trade",
+            "pl": "Handel wewnątrzkrajowy."
           }
         },
         {
-          "key": "export",
+          "code": "non-EU",
           "name": {
-            "en": "Export",
-            "pl": "Eksport"
+            "en": "non-EU",
+            "pl": "Poza UE"
+          },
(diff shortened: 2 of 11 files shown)


regimes/pl/extensions.go
```diff
@@ -34,34 +34,43 @@ var extensionKeys = []*cbc.KeyDefinition{
...
+				Code: "EU",
 				Name: i18n.String{
-					i18n.EN: "WDT",
-					i18n.PL: "WDT",
+					i18n.EN: "EU",
+					i18n.PL: "EU",
+				},
+				Desc: i18n.String{
+					i18n.EN: "Inter european trade",
+					i18n.PL: "Handel wewnątrzwspólnotowy.",
```
torrocus, line 50: It seems to me that the word "trade" is used less often in this context and it is more about "transactions". I was inspired by this [Polish website](https://www.biznes.gov.pl/en/publikacje/2274-transakcje-wewnatrzwspolnotowe).

```suggestion
					i18n.EN: "Intra-EU Transactions",
					i18n.PL: "Transakcje wewnątrzwspólnotowe",
```

I would wonder if this is a change in the right direction. I mean is the expansion of the shortcut:
> WDT - Wewnątrzwspólnotowa dostawa towarów

Maybe we should keep WDT and use the above description?

torrocus (commented):
Generally OK. It's worth considering the right names/descriptions.

regimes/pl/extensions.go
```diff
@@ -34,34 +34,43 @@ var extensionKeys = []*cbc.KeyDefinition{
...
+				Code: "non-EU",
 				Name: i18n.String{
-					i18n.EN: "Export",
-					i18n.PL: "Eksport",
+					i18n.EN: "non-EU",
+					i18n.PL: "Poza UE",
+				},
+				Desc: i18n.String{
+					i18n.EN: "Trading with countries outside the EU",
+					i18n.PL: "Handel z krajami z poza unii europejskiej.",
```
torrocus, line 72:
```suggestion
					i18n.PL: "Handel z krajami z poza Unii Europejskiej",
```

regimes/pl/extensions.go
```diff
@@ -34,34 +34,43 @@ var extensionKeys = []*cbc.KeyDefinition{
...
-				Key: "domestic",
+				Code: "domestic",
 				Name: i18n.String{
 					i18n.EN: "Domestic",
 					i18n.PL: "Krajowy",
 				},
-				// TODO: description required
+				Desc: i18n.String{
+					i18n.EN: "Domestic trade",
+					i18n.PL: "Handel wewnątrzkrajowy.",
```
torrocus, line 61:
```suggestion
					i18n.PL: "Handel wewnątrzkrajowy",
```

regimes/pl/extensions.go
torrocus, line 50:
```suggestion
					i18n.PL: "Handel wewnątrzwspólnotowy",
```

torrocus (approved):

regimes/pl/extensions.go
noplisu: I don't like the idea of keeping WDT because it is Polish specyfic shortcut and we would prefer to keep things more international. It is also specific for transporting goods where I want to make it more generic.

regimes/pl/extensions.go
```diff
@@ -34,34 +34,43 @@ var extensionKeys = []*cbc.KeyDefinition{
 		},
 	},
 	{
-		Key: ExtKeyKSeFVATZero,
+		Key: ExtKeyKSeFVATRegion,
```
samlown, line 37: Could this not be determined from the customer's country?

samlown (changes requested):
It strikes me that all of these things could be determined from the situation of the document... it might help to add examples.

regimes/pl/extensions.go
noplisu: Probably in most cases we could, but KSeF allows it not to be the case because it is more related to where the goods or services are provided than to who they are provided to.

torrocus (approved):

samlown:
Closing this one. Hopefully Poland will sort out what they want to do with electronic invoices eventually.

PR #269, opened 2024-04-16
PT: moving from region tags to extensions

samlown:
* Removed `azores` and `madeira` tags from PT region.
* Using extensions for regimes.
* Added support for `-` in `cbc.Code`.
* tax.Extensions provide extra utility functions.


cavalle (approved):
Changes LGTM! Since these changes would potentially allow different lines to use different tax regions per line, I confirmed that both InvoiceXpress and SAF-T supports that. So, all seems ok!

cavalle (approved):
Migration looks ok to me!

PR #270, opened 2024-04-17
MX: Stamps for SAT and CFDI

samlown:
* Adding stamps that will be required for PDF generation in Mexico.


cavalle (approved):
Makes sense to me 👍

samlown:
The objective here was to be able to automatically generate HTML/PDF versions of invoices just from the GOBL, but I'm now reaching the conclusion that copying all these extra fields does not make sense. The data is extensive, and the key concern is that they cannot actually by used to regenerate and validate the original XML. With the exception of the `sat-uuid` they are also not required for credit notes.

I'm thus concluding to discard these charges and instead adapt the github.com/invopop/gobl.html project to read the XML attachment and attempt to extract the details directly. The CFDI would be added to the PDF, and everything can be validated.

PR #274, opened 2024-04-23
UUID: moving to v7 and parse fix

samlown:
* Fixed parsing issue for empty UUIDs
* Moving to using UUID v7 as default.


cavalle (commented):
Very nice!

PR #279, opened 2024-04-26
`reverse-charge` changes and TicketBAI updates

samlown:
* The `reverse-charge` tag will no longer have an impact on tax calculations.
* ES: exemption reason updates, including those for reverse-charge.

### DIFF
--- a/regimes/es/extensions.go
+++ b/regimes/es/extensions.go
@@ -284,15 +284,43 @@ var extensionKeys = []*cbc.KeyDefinition{
 			{
 				Code: "OT",
 				Name: i18n.String{
-					i18n.EN: "Not subject: pursuant to Article 7 of the VAT Law. Other cases of non-subject.",
-					i18n.ES: "No sujeto: por el artículo 7 de la Ley del IVA. Otros supuestos de no sujeción.",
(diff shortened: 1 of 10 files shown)


regimes/es/extensions.go
```diff
@@ -284,15 +284,43 @@ var extensionKeys = []*cbc.KeyDefinition{
...
+			},
+			{
+				Code: "IE",
+				Name: i18n.String{
+					i18n.EN: "Not subject in the TAI due to localization rules, but foreign tax, IPS/IGIC or VAT from another EU member state is passed on",
+					i18n.ES: "No sujeto en el TAI por reglas de localización, pero repercute impuesto extranjero, IPS/IGIC o IVA de otro estado miembro UE",
+				},
+			},
+			{
+				Code: "S1",
```
samlown, line 313: @cavalle I'm wondering if this Code should be included at all... currently there for completeness.

samlown: Correct, but the key consideration is that you can't use more than one of them at the same time. They're mutually exclusive. Thinking of a UI selector, its easier if there is just one that covers all the cases. To that extent, you wouldn't expect to the see the `S1` option there... might be better to comment out for the time being.

PR #283, opened 2024-05-08
Refactor foreign tax identity handling

samlown:
* Ensures that no tax regime requires a Tax ID code to be present when validating the ID on its own.
* Fixes an issue detected when issuing an Invoice from Italy to France, and the French regime requiring the presence of a Tax ID code.


regimes/co/invoices.go
```diff
@@ -28,21 +28,28 @@ func (v *invoiceValidator) validate() error {
 				bill.InvoiceTypeCreditNote,
 				bill.InvoiceTypeProforma,
 			),
+			validation.Skip,
```
cavalle, line 31: It may not be obvious why this is necessary. For the benefit of future developers extending or looking at existing regimes to create new ones, would it be worth a comment explaining the need for the "skip"?

regimes/co/tax_identity.go
```diff
@@ -170,10 +170,9 @@ var taxIdentityTypeDefs = []*cbc.KeyDefinition{
 // validateTaxIdentity checks to ensure the NIT code looks okay.
 func validateTaxIdentity(tID *tax.Identity) error {
```
cavalle, line 171: Similar to my other comment, it may not be obvious that `validateTaxIdentity` functions aren't only called in the context of the regime they are defined in and so, they shouldn't validate anything that is not universal (like the tax code format). Would it be worth adding a comment at the top of each of these functions?

cavalle (approved):
Looks good to me! Just a couple of minor suggestion in the comments.

regimes/co/invoices.go
samlown: True, I thought about this, but there are soooo many Skips! I think this is just something that needs to be confirmed by someone with experience in the PRs. Some regional guidelines/README would help.

regimes/co/tax_identity.go
samlown: Same applies here, copying the same comments everywhere does scale all that well. A README guide for regimes would probably be useful.

PR #284, opened 2024-05-09
Exchange rates

samlown:
EPIC update here.
* GOBL now allows invoices to be generated in different currencies to that of the tax regime of the supplier.
* Validation made to ensure there is a conversion rate from document's currency to that of the regime.
* Support methods around exchange rates now included.


cavalle (approved):
Epic PR! Looks great to me!

PR #301, opened 2024-06-05
Maintaining accuracy on fuel balance totals

samlown:
* Fixing calculation issues for fuel balance totals by maintaining accuracy and removing support for fixing the tax amounts manually


cavalle (approved):
Oh yeah! Rock-solid precision 🔢

samlown:
> The last force-push is because I merged #302 unintentionally to this branch instead of main. I'm just leaving things as it were before that

That was actually intentional so that we don't get conflicts with the Changelog 😆

PR #302, opened 2024-06-10
Add `at-hash` stamp

cavalle:
* Related to: https://github.com/invopop/invoicexpress/pull/62


samlown (approved):
👍 Would also be nice to add and entry to the CHANGELOG.md, but I can do that once merged.

samlown:
> Thanks for the reminder. I need to get used to always update the changelog. I'll take care before merging

Might be a good idea to have some PR templates with checkboxes...

PR #307, opened 2024-06-20
Greece regime

cavalle:
* Adds Greece regime
* Covers the essential regime's config: additional extensions or taxes will be added in future PRs
* Not 100% sure about the approach took with the Islands Reduced VAT rates, please review my comments in the PR
* Closes APP-14

### DIFF
--- a/regimes/gr/README.md
+++ b/regimes/gr/README.md
@@ -1,0 +1,117 @@
+# 🇬🇷 GOBL Greece Tax Regime
+
+Greece uses the myDATA and Peppol BIS Billing 3.0 formats for their e-invoicing/tax-reporting system.
+
+Example GR GOBL files can be found in the [`examples`](./examples) (YAML uncalculated documents) and [`examples/out`](./examples/out) (JSON calculated envelopes) subdirectories.
+
+## Public Documentation
+
+* [myDATA API Documentation v1.0.7](https://www.aade.gr/sites/default/files/2023-10/myDATA%20API%20Documentation_v1.0.7_eng.pdf)
+* [Greek Peppol BIS Billing 3.0](https://www.gsis.gr/sites/default/files/eInvoice/Instructions%20to%20B2G%20Suppliers%20and%20certified%20PEPPOL%20Providers%20for%20the%20Greek%20PEPPOL%20BIS-EN-%20v1.0.pdf)
+* [VAT Rates](https://www.gov.gr/en/sdg/taxes/vat/general/basic-vat-rates)
+
+## Greece specifics
+
+### Islands' reduced tax rates
+
+Greece has three VAT rates: standard, reduced and super-reduced. These rates are reduced by a 30% on the islands of Leros, Lesbos, Kos, Samos and Chios.
+
+GOBL invoices including the `islands` tag inside the invoice's tax section will get the reduced rates applied instead of the regular ones. For example:
+
+```js
+{
+  "$schema": "https://gobl.org/draft-0/bill/invoice",
+  //...
+  "tax": {
+    "tags": [
+      "islands"
+    ]
+  },
+  // ...
+  "lines": [
+    {
+      "i": 1,
+      "quantity": "20",
+      "item": {
+        "name": "Υπηρεσίες Ανάπτυξης",
+        "price": "90.00",
+      },
+      "sum": "1800.00",
+      "taxes": [
+        {
+          "cat": "VAT",
+          "rate": "standard",
+          "percent": "17%" // Reduced island rate applied instead of the regular 24%
+        }
+      ],
+      "total": "1800.00"
+    }
+  ],
+}
+```
+
+### Tax exemptions
+
+Greece invoices can be exempt of VAT for different causes and the tax authority require a specific cause code to be provided.
+
+In a GOBL invoice, the `rate` of a line's tax need to be set to `exempt`, and the `ext` map's `gr-iapr-exemption` property needs to be set to one of these codes:
+
+| Code | Description                                             |
+| ---- | ------------------------------------------------------- |
+| `1`  | Without VAT - article 3 of the VAT code                 |
+| `2`  | Without VAT - article 5 of the VAT code                 |
+| `3`  | Without VAT - article 13 of the VAT code                |
+| `4`  | Without VAT - article 14 of the VAT code                |
+| `5`  | Without VAT - article 16 of the VAT code                |
+| `6`  | Without VAT - article 19 of the VAT code                |
+| `7`  | Without VAT - article 22 of the VAT code                |
+| `8`  | Without VAT - article 24 of the VAT code                |
+| `9`  | Without VAT - article 25 of the VAT code                |
+| `10` | Without VAT - article 26 of the VAT code                |
+| `11` | Without VAT - article 27 of the VAT code                |
+| `12` | Without VAT - article 27 - Seagoing Vessels of the VAT code |
+| `13` | Without VAT - article 27.1.γ - Seagoing Vessels of the VAT code |
+| `14` | Without VAT - article 28 of the VAT code                |
+| `15` | Without VAT - article 39 of the VAT code                |
+| `16` | Without VAT - article 39a of the VAT code               |
+| `17` | Without VAT - article 40 of the VAT code                |
+| `18` | Without VAT - article 41 of the VAT code                |
+| `19` | Without VAT - article 47 of the VAT code                |
+| `20` | VAT included - article 43 of the VAT code               |
+| `21` | VAT included - article 44 of the VAT code               |
+| `22` | VAT included - article 45 of the VAT code               |
+| `23` | VAT included - article 46 of the VAT code               |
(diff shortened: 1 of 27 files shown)


regimes/gr/README.md
```diff
@@ -0,0 +1,117 @@
...
+          "cat": "VAT",
+          "rate": "standard",
+          "percent": "17%" // Reduced island rate applied instead of the regular 24%
+        }
+      ],
+      "total": "1800.00"
+    }
+  ],
+}
+```
```
cavalle, line 51: I'm not 100% convinces about the approach for the alternative rates for some Greek islands. For two reasons:

1. In other countries we use a different approach for a similar problem: in [PT we use a tag extension](https://github.com/invopop/gobl/blob/main/regimes/pt/tax_categories.go#L46-L48) and [in MX we use a separated rate](https://github.com/invopop/gobl/blob/main/regimes/mx/tax_categories.go#L54-L55) for the border regions.
2. This solution forces us to [split the VAT Category Code](https://github.com/invopop/gobl/pull/307/files#diff-b511816f950af7d992d1ecabd067109dda861e5861765c0d5dd01dd417f609afR51-R52) that we'll need to provide to the IAPR.

An alternative approach would be to use separated rates but using combined keys [like we do in Spain for tax rate equivalence](https://github.com/invopop/gobl/blob/main/regimes/es/tax_categories.go#L82):
* `standard`
* `reduced`
* `super-reduced`
* `standard+islands`
* `reduced+islands`
* `super-reduced+islands`

This solution would let us avoid the IAPR vat category split mentioned in point 2 above.

BTW, `islands` is the best code/tag I've been able to come up with, but I'm open to suggestions here too.

samlown: Hehe nice. So the key question is, ¿do we need to identify the different rates like in Portugal? If not, then I'd be tempted to not worry about this and just let the client put whatever rates they like in for the exception cases.

If rate keys or any other data are required, then I'd consider going with extensions with a code for the region that has a reduced rate.

As you say, `islands` is too generic and doesn't really give enough information about where. Nor does it deal with future exception cases.

samlown: So I see there is indeed a code for the VAT category with 6 options in total. I think the best bet is to add this property as an extension. We want to avoid using tags at the document level as they don't account for mixing rates in the same document and don't coincide with how most ERPs deal with tax on a per-line basis.

We may need to write some extra code to handle these cases though so that the extension is added automatically when a rate is provided, and validated to ensure it doesn't conflict... I'll add some more comments in-line.

regimes/gr/extensions.go
```diff
@@ -0,0 +1,306 @@
...
+		Key: ExtKeyIAPRVATCat,
+		Name: i18n.String{
+			i18n.EN: "VAT category",
+			i18n.EL: "Κατηγορία ΦΠΑ",
+		},
+		Codes: []*cbc.CodeDefinition{
+			{
+				Code: "1",
+				Name: i18n.String{
+					i18n.EN: "VAT rate 24%",
```
samlown, line 25: Should numbers be listed in the name? I'd go with the standard names I think and avoid numbers to reduce issues when they change.

regimes/gr/README.md
```diff
@@ -0,0 +1,134 @@
...
+* [Greek Peppol BIS Billing 3.0](https://www.gsis.gr/sites/default/files/eInvoice/Instructions%20to%20B2G%20Suppliers%20and%20certified%20PEPPOL%20Providers%20for%20the%20Greek%20PEPPOL%20BIS-EN-%20v1.0.pdf)
+* [VAT Rates](https://www.gov.gr/en/sdg/taxes/vat/general/basic-vat-rates)
+
+## Greece specifics
+
+### VAT categories
+
+Greece has three VAT rates: standard, reduced and super-reduced. Each of these rates are reduced by a 30% on the islands of Leros, Lesbos, Kos, Samos and Chios. The tax authority identifies each rate with a specific VAT category.
+
+In GOBL, the IAPR VAT category code must be set using the `iapr-gr-vat-category` extension of a line's tax to one of these codes:
```
samlown, line 19:
```suggestion
In GOBL, the IAPR VAT category code must be set using the `iapr-gr-vat-cat` extension of a line's tax to one of these codes:
```

regimes/gr/gr.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+func init() {
+	tax.RegisterRegime(New())
+}
+
+// Custom keys used typically in meta or codes information.
+const (
+	KeyIAPRVATCategory cbc.Key = "iapr-vat-category"
+)
```
samlown, line 21: Comments below, but I think this one should be removed and we can use the extension key.

regimes/gr/tax_combo.go
```diff
@@ -0,0 +1,46 @@
...
+	if tc.Ext == nil {
+		tc.Ext = make(tax.Extensions)
+	}
+
+	if tc.Ext.Has(ExtKeyIAPRVATCat) {
+		return nil
+	}
+
+	vcat := rate.Map[KeyIAPRVATCategory]
+	tc.Ext[ExtKeyIAPRVATCat] = tax.ExtValue(vcat)
```
samlown, line 43: This feels a bit strange with the two types of keys. Perhaps we can use `ExtKeyIAPRVatCat` inside the Rate Codes list definitions directly, so that it is very obvious we're talking about the same key?

tax/set.go
```diff
@@ -102,6 +102,10 @@ func combineExtKeys(cat *Category, rate *Rate) []cbc.Key {
 // prepare updates the Combo object's Percent and Retained properties using the base totals
 // as a source of additional data for making decisions.
 func (c *Combo) prepare(r *Regime, tags []cbc.Key, date cal.Date) error {
+	if err := r.CalculateObject(c); err != nil {
+		return err
+	}
```
samlown, line 107: Did you test this at the bottom already? While not the case for GR, other countries may need some of the other calculations to be performed beforehand.

regimes/gr/tax_categories.go
```diff
@@ -0,0 +1,99 @@
...
+				},
+				Values: []*tax.RateValue{
+					{
+						Percent: num.MakePercentage(6, 2),
+					},
+				},
+				Map: cbc.CodeMap{
+					KeyIAPRVATCategory: "3",
+				},
+			},
```
samlown, line 82: We lost the island rates here... I do think they can still be supported using the `standard+island` (`tax.RateStandard.With(RateIsland)`) format in the Key.

samlown (changes requested):
This looks really good! A few comments to consider, but from the outset I'm feeling very happy with this approach!

regimes/gr/tax_categories.go
cavalle: I thought they weren't essential at this stage, and I wasn't sure about the `+` idea. I'll add them!

tax/set.go
cavalle: I _think_ I saw that, for other structs, the regimes calculation are executed _before_ the GOBL ones. But I'll double-check.

regimes/gr/tax_combo.go
cavalle: I'll use the same key :+1:

regimes/gr/extensions.go
cavalle: I used the same descriptions that appear in the myDATA docs (both in Greek and English versions) for consistency with the specs. But it's true those will become outdated. I'll rename them to use the rate names.

tax/set.go
samlown: It might change according to the model... in this case there are preparations so it probably makes sense lower down.

samlown (approved):
🔝  very nice.

regimes/gr/extensions.go
cavalle: Addressed in [512ab72](https://github.com/invopop/gobl/pull/307/commits/512ab72a85133320e6986ac3f1e59dc2fab0285c)

regimes/gr/README.md
cavalle: Addressed in [512ab72](https://github.com/invopop/gobl/pull/307/commits/512ab72a85133320e6986ac3f1e59dc2fab0285c)

regimes/gr/gr.go
cavalle: Addressed ins [512ab72](https://github.com/invopop/gobl/pull/307/commits/512ab72a85133320e6986ac3f1e59dc2fab0285c)

regimes/gr/tax_combo.go
cavalle: Addressed in [512ab72](https://github.com/invopop/gobl/pull/307/commits/512ab72a85133320e6986ac3f1e59dc2fab0285c)

tax/set.go
cavalle: Agreed. I took care of it in [34df770](https://github.com/invopop/gobl/pull/307/commits/34df77054406dfea18c2e47471eff385cb992570)

PR #317, opened 2024-07-18
bill.Invoice: convert into a new currency

samlown:
* Adding support for a new `bill.Invoice` `ConvertInto(cur)` method that will convert all of an invoices amounts into the defined currency.

### DIFF
--- a/bill/invoice_convert.go
+++ b/bill/invoice_convert.go
@@ -1,0 +1,121 @@
+package bill
+
+import (
+	"fmt"
+
+	"github.com/invopop/gobl/currency"
+	"github.com/invopop/gobl/pay"
+)
+
+// ConvertInto will use the defined exchange rates in the invoice to convert all the prices
+// into the given currency.
+//
+// The intent of this method is help convert the invoice amounts when the destination is
+// unable or unwilling to handle the current currency. This is typically the case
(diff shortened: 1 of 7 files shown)


bill/invoice_convert.go
```diff
@@ -0,0 +1,173 @@
...
+	i2 := *inv
+	i2.Totals = new(Totals)
+	i2.Lines = inv.convertLines(ex)
+	i2.Discounts = inv.convertDiscounts(ex)
+	i2.Charges = inv.convertCharges(ex)
+	i2.Outlays = inv.convertOutlays(ex)
+	i2.Payment = inv.convertPayment(ex)
+	i2.Currency = cur
+
+	if err := i2.Calculate(); err != nil {
```
cavalle, line 47: I think the approach makes sense, first convert the amounts in the invoice, and the calculate again. This will make the totals consistent with the rest of the invoice (i.e. not rounding discrepancies)

cavalle (approved):
LGTM! (at least, as a first version until we can test it in the wild)

bill/invoice_convert.go
samlown: Also, there are a large number of amounts to convert! 😆

PR #318, opened 2024-07-19
DE: freelance tax number support

samlown:
* Invoices can now be issued with either a Tax ID Code, *or* the German "Tax Number" identity.
* Simplified Invoices in Germany remove the requirement for the presence of any supplier tax code.


cavalle (approved):
LGTM

PR #299, opened 2024-05-30
Preparing ground work for new bill.Payment schema

samlown:

PR #322, opened 2024-07-22
WIP: introducing concept of `bill.Receipt`

samlown:
* Refining concepts around `bill.Receipt` to use for payment and remittence documents.
* Replaces #299


samlown:
Replaced with #379

PR #323, opened 2024-07-22
CO: debit notes (II)

cavalle:
* Extends #320 
* Part of APP-50

### DIFF
--- a/regimes/co/extensions.go
+++ b/regimes/co/extensions.go
@@ -8,9 +8,9 @@ import (
 
 // Extension keys used in Colombia.
 const (
-	ExtKeyDIANMunicipality     cbc.Key = "co-dian-municipality"
-	ExtKeyDIANCorrectionCredit cbc.Key = "co-dian-correction"
-	ExtKeyDIANCorrectionDebit  cbc.Key = "co-dian-correction-debit"
+	ExtKeyDIANMunicipality cbc.Key = "co-dian-municipality"
+	ExtKeyDIANCreditCode   cbc.Key = "co-dian-credit-code"
+	ExtKeyDIANDebitCode    cbc.Key = "co-dian-debit-code"
 )
 
 var extensionKeys = []*cbc.KeyDefinition{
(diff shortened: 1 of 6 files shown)


regimes/co/extensions.go
```diff
@@ -8,8 +8,9 @@ import (
 
 // Extension keys used in Colombia.
 const (
-	ExtKeyDIANMunicipality cbc.Key = "co-dian-municipality"
-	ExtKeyDIANCorrection   cbc.Key = "co-dian-correction"
+	ExtKeyDIANMunicipality     cbc.Key = "co-dian-municipality"
+	ExtKeyDIANCorrectionCredit cbc.Key = "co-dian-correction"
```
cavalle, line 12: We need to separated correction extensions now, one for credit notes and another one for debit notes. I was shy to rename `co-dian-correction` to `co-dian-correction-credit`, but it's probably ok, nothin will break if we do that. @samlown WDYT?

regimes/co/extensions.go
```diff
@@ -91,4 +92,41 @@ var extensionKeys = []*cbc.KeyDefinition{
...
+				Name: i18n.String{
+					i18n.EN: "Interest",
+					i18n.ES: "Intereses",
+				},
+			},
+			{
+				Code: "2",
+				Name: i18n.String{
+					i18n.EN: "Pending charges",
+					i18n.ES: "Gastos por cobrar",
```
cavalle, line 113: The Spanish names are those in the DIAN docs. I'm not 100% sure about the English translation in this particular case.

regimes/co/extensions.go
samlown: oh, that's very annoying... Do you know what the underlying code is called? This key was written some time ago, I'd be more tempted now to use a directly translation of the XML field name.

cavalle: These are the docs regarding those fields:

image

In both cases, the same UBL field is used, but the list of allowed values and their meaning are different depending on whether the type is a debit or a credit note. That's why I split the extensions in two.

samlown: I think I'd be tempted to go with simply:
* `co-dian-credit-code`
* `co-dian-debit-code`
Reasoning: "correction" is implied in "credit" and "debit", and "code" makes it clear that this is something defined in a list of codes somewhere.

cavalle: Updating!

cavalle: Address in [88db17e](https://github.com/invopop/gobl/pull/323/commits/88db17e75e12fb4bd503cb91cf4cefa0763fafb8)

regimes/co/extensions.go
samlown: Translation seems good to me.

samlown (approved):
Looks great. I like the names to, "correction" felt too generic

PR #325, opened 2024-07-22
Renaming Greek country code to `EL`

samlown:
* It turns out Greece does not recognise `GR` for tax codes and instead uses `EL`.
* Define 4 new country codes to use for these special tax exception cases.


Menendez6 (approved):

cavalle (approved):
Let's do it!

PR #330, opened 2024-07-24
Add debit notes and validations to PT regime

cavalle:


samlown:
@cavalle just to double check, are you sure that debit notes are supported in Portugal?

samlown (approved):

PR #333, opened 2024-07-24
Support for Header links

samlown:


cavalle (approved):
Not sure what's the use case for this feature. But I'm sure I'll find out eventually :D The changes look good.

samlown:
> Not sure what's the use case for this feature. But I'm sure I'll find out eventually :D The changes look good.

😆  its for the new Portal service. Instead of adding links to meta rows inside Invopop, I felt it made a lot of sense to include them in the stamps so that other apps, such as the PDF generator, can use them. It may also be useful as an alternative to stamps for specific URLs, like QR codes.

PR #339, opened 2024-07-30
Add support for simplified invoices in Greece

cavalle:
* Tunes validations to handle simplified invoices in Greece
* Part of APP-17

### DIFF
--- a/regimes/gr/invoices.go
+++ b/regimes/gr/invoices.go
@@ -9,34 +9,47 @@ import (
 	"github.com/invopop/validation"
 )
 
+// invoiceValidator adds validation checks to v.invoices which are relevant
+// for the region.
+type invoiceValidator struct {
+	inv *bill.Invoice
+}
+
 func validateInvoice(inv *bill.Invoice) error {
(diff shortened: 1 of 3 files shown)


regimes/gr/invoices.go
```diff
@@ -9,34 +9,47 @@ import (
 	"github.com/invopop/validation"
 )
 
+// invoiceValidator adds validation checks to v.invoices which are relevant
+// for the region.
+type invoiceValidator struct {
```
samlown, line 14: I'd been refactoring these structs out. They didn't seem to add value.

A more functional approach might be to return methods with the specific conditions that will be checked for. This is a little lighter than adding things to the heap, and has a more explicit interface:

For example
```go
func validateInvoiceParty(tags []cbc.Key) validation.RuleFunc {
  return func(value any) error {
    // type checks
    if tax.TagSimplified.In(tags) {
      // validate simplified
    } else {
      // validate normal
    }
  }
}
```

samlown (approved):
LG, but I'd consider a more functional approach.

PR #347, opened 2024-08-26
Support specific country code inside tax combos

samlown:
* This large PR ensures that the `customer-rates` tag included in an invoice will copy the customer's tax ID country to each individual tax row, thus ensuring that the customer's tax rates are applied.
* In order to support this, it made sense to remove the restrictions around tax regimes, its now possible to create a GOBL invoice for any country in the world, albeit with reduced validation processes.


regimes/it/extensions.go
```diff
@@ -58,8 +58,13 @@ var extensionKeys = []*cbc.KeyDefinition{
...
 		Desc: i18n.String{
 			i18n.EN: here.Doc(`
+<<<<<<< HEAD
 				Code used to describe the type of document being sent to the SDI. This should
 				normally be determined automatically by GOBL.
+=======
+				Code used to describe the type of document being sent to the SDI. This is
+				used to determine the correct schema to use when validating the document.
+>>>>>>> 1aca6b8 (Refactor scenarios to apply extensions to document automatically)
 			`),
```
cavalle, line 68: You left this unsolved conflict. Interestingly, it's valid Go code!

cavalle (changes requested):
Changes look good to me. May not be the final destination regarding customer rates but definitely feels like a step in the right direction.

I just pointed out an unresolved conflict in the code that needs to be fixed.

regimes/it/extensions.go
samlown: ooops! Good catch! Perhaps the editor didn't show it as it was inside a text block...

PR #350, opened 2024-08-29
Remove draft flag

samlown:
After much debate, this PR finally removes the "draft" boolean property from the envelope headers.

Draft status was previously used to apply additional validation rules to header stamps and special fields like the invoice "code". This functionality has now been moved to the signatures. If an envelope is signed, thus locking in the potential for any changes, the additional validation rules will be run.

This change comes after continued confusion about the role and usefulness of the `draft` flag. The signature functionality has always been in GOBL and is far easier to understand, so this is a natural replacement.


samlown:
Closing in favour of #359

PR #353, opened 2024-09-06
Income classification extensions for GR

cavalle:


cbc/value_definition.go
```diff
@@ -0,0 +1,29 @@
+package cbc
+
+import (
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/validation"
+)
+
+// ValueDefinition describes a specific value and how it maps to a human name
+// and description if appropriate.
+type ValueDefinition struct {
```
samlown, line 10: Seeing it here, I'm just not feeling the love for `ValueDefinition` alongside `CodeDefinition` and `KeyDefinition`... it feels overloaded and confusing. I'm wondering if there is space for a general purpose `EnumDefinition` (or could still be `ValueDefinition`) that replaces the use-cases of Keys, Codes, and Strings embedded inside another `KeyDefinition`.

I'll do a quick experiment with a fork of this branch.

PR #354, opened 2024-09-09
Refactoring Key, Code, and Value definitions to single approach

samlown:
* Moving the list of possible values from keys and codes to a single "value" definition.


cavalle (approved):
I wholeheartedly approve! 💯

PR #366, opened 2024-09-23
Add default scenario for mydata standard invoices

cavalle:


samlown (approved):
After discussion and the new `other` invoice type, this all looks good.

PR #369, opened 2024-09-23
More flexible cbc.Code, better normalization

samlown:
* Finally allowing spaces, underscores, and lower-case letters inside `cbc.Code` given the usage with invoice series and code.
* Normalization will now clean multiple separator symbols to use the first symbol found, which is the "essence" of Codes.


cavalle (approved):
LGTM!

PR #370, opened 2024-09-24
Australian Tax Regime

apardods:
Added support for Australian regime. Key considerations:
- Some sources allow ACN (Australian Company Number) to be used as ID instead of the usual ABN (Australian Business Number). As most official sources disallow it this support is removed, but since every ACN is uniquely linked to an ABN, it is not fully removed but commented out.
- Examples are not yet included.
- The Wine Equalisation Tax and Luxury Car Tax are treated as scenarios, but they could also potentially be considered categories, as they are separate taxes with a fixed rate. [More Info](https://taxsummaries.pwc.com/australia/corporate/other-taxes).
- Some countries include their income taxes in categories, while others do not. Taking into account the invoicing scope of the regime, income taxes are not included in Australian support.
- The extensions that other countries include in the addons folder are not included, as they fall into the PEPPOL integration. Therefore the invoicing logic that could be moved there is kept in the regime folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Added
+
+- `gr-mydata`: added `gr-mydata-other-tax` extension to set the category of other taxes in charges.
+
 ## [v0.207.0] - 2024-12-12
 
 ### Added
--- a/addons/gr/mydata/extensions.go
+++ b/addons/gr/mydata/extensions.go
@@ -13,8 +13,7 @@ const (
 	ExtKeyIncomeCat    = "gr-mydata-income-cat"
 	ExtKeyIncomeType   = "gr-mydata-income-type"
 	ExtKeyPaymentMeans = "gr-mydata-payment-means"
-
-	InvoiceTypeRetailPrefix = "11."
+	ExtKeyOtherTax     = "gr-mydata-other-tax"
 )
 
 var extensions = []*cbc.Definition{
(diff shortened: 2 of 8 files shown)


regimes/au/au.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/tax"
+)
+
+func init() {
+	tax.RegisterRegimeDef(New())
+}
+
+const (
+	TaxCategoryWET cbc.Code = "WET"
```
samlown, line 17: Constants also need comments, either for the comment block or the specific example. You can run the linter locally (`golanglint-ci`), or see the Github suggestions.

regimes/au/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+	return validation.ValidateStruct(inv,
+		validation.Field(&inv.Type,
+			validation.In(
+				bill.InvoiceTypeStandard,
+				bill.InvoiceTypeCreditNote,
+				bill.InvoiceTypeDebitNote,
+			),
+			validation.Skip,
+		),
+		validation.Field(&inv.IssueDate,
```
samlown, line 32: Issue date is always required by default, so this is not required.

regimes/au/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+func validateInvoice(inv *bill.Invoice) error {
+	v := &invoiceValidator{inv: inv}
+	return v.validate()
+}
+
+func (v *invoiceValidator) validate() error {
+	inv := v.inv
+	return validation.ValidateStruct(inv,
+		validation.Field(&inv.Type,
+			validation.In(
```
samlown, line 25: Type validation isn't normally required either, we've found this to be a bit too restrictive.

regimes/au/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+			validation.Skip,
+		),
+		validation.Field(&inv.Supplier,
+			validation.By(v.supplier),
+			validation.Skip,
+		),
+		// Customer ID necessary when total over AUD 1000
+		validation.Field(&inv.Customer,
+			validation.When(
+				inv.Totals.Total.Compare(num.MakeAmount(1000, 0)) == 1,
```
samlown, line 43: This is a great idea, but a few caveats which mean its best avoided. We haven't added any of these types of amount based validations as they change over time. The only way to implement it correctly would be with some kind of time database. The other issue is that there are *always* exceptions, its very hard to generalise, so its best to avoid.

In terms of technical implementation, there are a few gotchas:
* the `num.MakeAmount(1000, 0)` ideally should be a constant somewhere.
* if the `Totals` or `Total` values are nil, this will cause a panic. Unlikely, but this is validation, so anything can happen!

regimes/au/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/l10n"
+	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/gobl/tax"
+	"github.com/invopop/validation"
+)
+
+type invoiceValidator struct {
```
samlown, line 12: This is fine, but we're moving away from using validator structs to just have method calls. There isn't much value in the struct in the end.

regimes/au/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (v *invoiceValidator) supplier(val any) error {
+	obj, _ := val.(*org.Party)
+	if obj == nil {
+		return nil
+	}
+	return validation.ValidateStruct(obj,
+		validation.Field(&obj.TaxID,
+			validation.Required,
+			tax.RequireIdentityCode,
```
samlown, line 59: Something to validate. Do all invoices in AU require an Identity Code? They certainly do in countries like Spain, but many other countries around the world have limits that mean you can issue invoices without being registered for tax, and thus don't need an ID.

Given the comment in the README at the beginning regarding ABN and ACN, I'd expect there to be some some special situations to deal with.

An important consideration here which we incorporated into GOBL recently, is that the Tax ID should always be an ID that can be used internationally. I'm not sure what the situation is in AU, but there is usually only one business code that is valid for import & export.

regimes/au/tax_categories.go
```diff
@@ -0,0 +1,111 @@
...
+			i18n.EN: "Goods and Services Tax",
+		},
+		Retained: false,
+		Rates: []*tax.RateDef{
+			{
+				Key: tax.RateExempt,
+				Name: i18n.String{
+					i18n.EN: "GST-free",
+				},
+				Values: []*tax.RateValueDef{
```
samlown, line 28: Values aren't needed for "Exempt" rates. You can use the `Exempt: true` flag instead of values.

regimes/au/tax_categories.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		Title: i18n.String{
+			i18n.EN: "Goods and Services Tax",
+		},
+		Retained: false,
+		Rates: []*tax.RateDef{
+			{
+				Key: tax.RateExempt,
+				Name: i18n.String{
+					i18n.EN: "GST-free",
```
samlown, line 26: I'd keep names simple:
```suggestion
					i18n.EN: "Exempt",
```

regimes/au/tax_categories.go
```diff
@@ -0,0 +1,111 @@
...
+					{
+						Since:   cal.NewDate(2000, time.July, 1),
+						Percent: num.MakePercentage(10, 2),
+					},
+				},
+			},
+			{
+				Key: tax.RateZero,
+				Name: i18n.String{
+					i18n.EN: "Input-Taxed",
```
samlown, line 50: I'd just use "Zero" for the name here.

regimes/au/tax_categories.go
```diff
@@ -0,0 +1,111 @@
...
+
+var taxCategories = []*tax.CategoryDef{
+	{
+		Code: tax.CategoryGST,
+		Name: i18n.String{
+			i18n.EN: "GST",
+		},
+		Title: i18n.String{
+			i18n.EN: "Goods and Services Tax",
+		},
```
samlown, line 20: It'd be useful to incorporate the "sources" here so that we know where the data came from.

regimes/au/tax_categories.go
```diff
@@ -0,0 +1,111 @@
...
+				Values: []*tax.RateValueDef{
+					{
+						Since:   cal.NewDate(2000, time.July, 1),
+						Percent: num.MakePercentage(0, 2),
+					},
+				},
+			},
+		},
+	},
+	//Source: https://www.ato.gov.au/businesses-and-organisations/gst-excise-and-indirect-taxes/wine-equalisation-tax
```
samlown, line 61: This is good, but you can add it to the raw data in the "Source" attribute. The `fr` regime has an example of this.

regimes/au/tax_identity.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		sum += digit * taxWeightTableABN[i]
+	}
+	if sum%89 == 0 {
+		return nil
+	}
+	return errors.New("checksum mismatch")
+}
+
+// func companyNumberCheck(val string) error {
```
samlown, line 69: As a rule, its best to remove commented blocks from PRs. This however does look useful, so I'd encourage a deeper look into how the ABN and ACN should be used.

See the `DE` regime for an example of two types of codes being used.

regimes/gb/tax_categories.go
```diff
@@ -39,7 +41,7 @@ var taxCategories = []*tax.CategoryDef{
 				},
 				Values: []*tax.RateValueDef{
 					{
-						Since:   cal.NewDate(2011, 1, 4),
+						Since:   cal.NewDate(2011, time.April, 1),
```
samlown, line 44: Nice catch :-)

samlown (changes requested):
In general, this is a great start! We do like READMEs also. I'd consider adding more links to sources of information though. ChatGPT can also help with refining the text and examples.

There are a few in-line comments.

I'd encourage you to figure out what the situation with the Tax IDs is in AU, as its very likely both will need to be supported.

apardods:
Thanks for the feedback! Upon further research, I believe ACNs should not be supported, as all official sources only mention the use of the ABN. I initially included since to have an ACN a business must previously have an ABN, but for invoicing purposes it seems like ABN is the key identifier.

As for requiring an Identity Code, the law says it is necessary for all totals over AUD 82.50 as mentioned [here](https://www.ato.gov.au/businesses-and-organisations/gst-excise-and-indirect-taxes/gst/tax-invoices), and since I also removed the totals check I'm not sure I should add that step.

I have also implemented the other minor changes in the comments, and if there are any changes to be reverted (especially regarding tax IDs) I'd be happy to bring them back!

samlown:
> I believe ACNs should not be supported, as all official sources only mention the use of the ABN. I initially included since to have an ACN a business must previously have an ABN, but for invoicing purposes it seems like ABN is the key identifier.

Are you sure? This is what ChatGPT said:

When is the ACN Required?

 1. If the entity is a company and does not include the ABN on the invoice:
   * If your company does not use an Australian Business Number (ABN), you must include the ACN on documents such as invoices, receipts, and other official paperwork.
 2. If you wish to display both:
   * Some businesses include both the ACN and ABN for clarity, especially if they operate in contexts where company identification is important.

apardods:
I think the reason why only ABN shows up in these guides and is treated as the default is because ACN seems to be reserved for larger companies where invoicing is typically taken care of. Since an ACN is linked directly to an ABN, I have added back support for the ACN.

The reason I removed it was that it only appeared in one or two unofficial sources and I was not convinced, but with the context of it being reserved for larger companies and being linked directly to an ABN, it makes sense to include it. If there's anything else you would like included let me know!

regimes/au/tax_categories.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+		Title: i18n.String{
+			i18n.EN: "Goods and Services Tax",
+		},
+		Sources: []*tax.Source{
+			{
+				Title: i18n.String{
+					i18n.EN: "How GST Works",
+				},
+				URL: "https://www.ato.gov.au/businesses-and-organisations/gst-excise-and-indirect-taxes/gst/how-gst-works",
```
samlown, line 26: I couldn't find the definitions of the actual tax rates here... the objective here is to have a specific page that can be used to determine what the latest rates are.

samlown:
@apardods it looks like the branch for this PR got force overwritten :-( do you still have the original branch on your local machine or somewhere else? Perhaps you can recover the changes with the reflog command? https://github.com/orgs/community/discussions/64693 Our bad for not merging this one soon enough as there were still a few things to sort out.

PR #378, opened 2024-10-01
Added Support for Steuerliche Identifikationsnummer and Validation for Steuernummer

apardods:
In Germany, individuals can invoice internally using the usual Steuernummer, for which testing was added, or the new Identifikationsnummer (or tax ID). Support for this value was added while both are in use. The latter is expected to phase out the former.


samlown:
@apardods is it safe to assume this is replaced with #381 ?

PR #382, opened 2024-10-04
PT: Updating reduced rate for PT-MA

tomasguinzburg:
Updated reduced VAT rate for Madeira [from 5% to 4%](https://invoicexpress.com/nova-taxa-reduzida-iva-madeira/). 

Tax categories ordering is not validated for categories with extensions. From testing the example and checking the code my understanding is that the most current version of a tax rate should go first.


samlown (approved):
Many thanks for this! There are a few lint issues, but I'll handle those 👍

PR #386, opened 2024-10-04
CO: moving DIAN UBL requirements to Addon

samlown:


cavalle (approved):
LGTM! I couldn't spot any rename or refactor that should break anything in the provider, but please let me know if I missed anything

PR #388, opened 2024-10-07
improving documentation and examples for IT an MX

juanmoliner:


samlown (approved):
👍

samlown (changes requested):
Just realised this is missing some files. The output for the `it` example, and the input for the `mx` example.

PR #393, opened 2024-10-15
X-Rechnung Add-on

apardods:
* Added `de-xrechnung-v3` addon.
* Added `eu-en16931-v2017` addon that handles universal normalization and validation required for compliance.
* Added `iso` and `untdid` "catalogues" used for storing universally available extensions.
* Removed Payment UNTDID references, now supported by catalogues.


addons/de/xrechnung/extensions.go
```diff
@@ -0,0 +1,84 @@
+package xrechnung
+
+import (
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+)
+
+// ExtKeyTaxRate is the key for the tax rate extension in XRechnung
+var ExtKeyTaxRate cbc.Key = "de-xrechnung-tax-rate"
```
samlown, line 9: Is "tax rate" the correct name? The definitions also cover tax categories. I'd strongly suggest adding a comment with the link to the source information to be able to double check.

addons/de/xrechnung/instructions.go
```diff
@@ -0,0 +1,107 @@
...
+var validPaymentKeys = []cbc.Key{
+	pay.MeansKeyCash,
+	pay.MeansKeyCheque,
+	pay.MeansKeyCreditTransfer,
+	pay.MeansKeyCard,
+	pay.MeansKeyDirectDebit,
+	pay.MeansKeyOther,
+	KeyPaymentMeansSEPACreditTransfer,
+	KeyPaymentMeansSEPADirectDebit,
+}
```
samlown, line 25: Please check the other regime and addon implementations. It is important that the base GOBL types are maintained and only extended if needed with the `+` stuff.

addons/de/xrechnung/instructions.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ValidatePaymentInstructions validates the payment instructions according to the XRechnung standard
+func ValidatePaymentInstructions(value interface{}) error {
```
samlown, line 28: I'd keep these methods private, i.e. lower-case `v` at the start.

addons/de/xrechnung/instructions.go
```diff
@@ -0,0 +1,107 @@
...
+	inv, ok := value.(*bill.Invoice)
+	if !ok || inv == nil || inv.Payment == nil || inv.Payment.Instructions == nil {
```
samlown, line 30: This doesn't look right. I'd expect the `ValidatePaymentInstructions` method to only validate a `pay.Instructions` instance.

addons/de/xrechnung/instructions.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil
+	}
+	instr := inv.Payment.Instructions
+	return validation.ValidateStruct(instr,
+		validation.Field(&instr.Key,
+			validation.Required,
+			validation.By(validatePaymentKey),
```
samlown, line 37: missing `validation.Skip` under here (there are lots of other cases also)

addons/de/xrechnung/instructions.go
```diff
@@ -0,0 +1,107 @@
...
+				validation.By(validateDirectDebit),
+			),
+		),
+	)
+}
+
+func validatePaymentKey(value interface{}) error {
+	t, ok := value.(cbc.Key)
+	if !ok {
+		return validation.NewError("type", "Invalid payment key")
```
samlown, line 65: I'm not sure about this usage of `NewError`, what is "`type`"? Also, convention is that error messages should not start with capital letters.

addons/de/xrechnung/instructions.go
```diff
@@ -0,0 +1,107 @@
...
+
+func validateDirectDebit(value interface{}) error {
+	dd, ok := value.(*pay.DirectDebit)
+	if !ok || dd == nil {
+		return nil
+	}
+	return validation.ValidateStruct(dd,
+		// BR-DE-29 - Changed to Peppol-EN16931-R061
+		validation.Field(&dd.Ref,
+			validation.Required.Error("Mandate reference is mandatory for direct debit"),
```
samlown, line 81: Are these messages necessary?

addons/de/xrechnung/instructions_test.go
```diff
@@ -0,0 +1,100 @@
...
+					CreditTransfer: []*pay.CreditTransfer{
+						{
+							IBAN: "DE89370400440532013000",
+							BIC:  "DEUTDEFF",
+						},
+					},
+				},
+			},
+		}
+		assert.NoError(t, xrechnung.ValidatePaymentInstructions(inv))
```
samlown, line 28: I'd run the validation on the whole Invoice to make sure nothin else is broken. Please see the other examples.

addons/de/xrechnung/invoices.go
```diff
@@ -0,0 +1,198 @@
...
+var validTypes = []cbc.Key{
+	bill.InvoiceTypeStandard,
+	bill.InvoiceTypeCreditNote,
+	bill.InvoiceTypeCorrective,
+	invoiceTypeSelfBilled,
+	invoiceTypePartial,
+	invoiceTypePartialConstruction,
+	invoiceTypePartialFinalConstruction,
+	invoiceTypeFinalConstruction,
+}
```
samlown, line 27: Rather than modify the invoice types (which is strictly prohibited), checkout the usage of "tags" and "scenarios" in other addons.

samlown (changes requested):
Its going in the right direction, but lot's of changes required. Please do closely look at the other example addons to make sure everything is as similar as possible. Take special note of the usage of `validation.Skip` for example also, this is important.

cavalle (approved):
Beautiful! Totally feels like the right way to separate the extensions, scenarios, validations and normalisations derived from EN16931 from those derived from each specific CIUS/Extension.

PR #394, opened 2024-10-16
EN 16931 tags

apardods:
Added tags for GOBL objects mapping them to their correspondent EN 16931 Business Term (BT) or Business Group (BG).

### DIFF
--- a/bill/charges.go
+++ b/bill/charges.go
@@ -16,13 +16,13 @@ import (
 // TODO: use UNTDID 7161 code list
 type LineCharge struct {
 	// Percentage if fixed amount not applied
-	Percent *num.Percentage `json:"percent,omitempty" jsonschema:"title=Percent"`
+	Percent *num.Percentage `json:"percent,omitempty" jsonschema:"title=Percent" en16931:"BT-143"`
 	// Fixed or resulting charge amount to apply (calculated if percent present).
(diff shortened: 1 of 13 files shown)


cavalle (approved):
Nice!! I had exactly the same idea a few days ago

samlown (changes requested):
The code itself looks sound, but I'm afraid this isn't the best approach. The fundamental issue is that the EN16931 semantic codes usually have a context, whereas the syntax defined in the GOBL structures are designed to be used anywhere.

The objective however is sound, its just the approach that needs changing. In my opinion there should be a table of EN16931 codes to JSON pointers/paths of attributes inside GOBL structures so that you always have a context. This table should also define "notes" or something similar to describe any additional implications or meanings. The mapping table should also be adjusted when addons or regimes are applied.

This might be a good subject to brainstorm next week.

cavalle:
It's a good point that GOBL structs have a scope that could potentially exceed that of EN16391 and putting the metadata directly in the structs, while convenient, may not makes any sense in other contexts.

It's also true that mapping may not always be direct or trivial and additional notes to indicate special details may also be useful or necessary.

urakozz:
Thank you for this PR, it's more comprehensive than FacturX documentation 😄

samlown:
Closing this one in favor of #401

PR #396, opened 2024-10-18
Brazil regime

cavalle:


regimes/br/invoices.go
```diff
@@ -0,0 +1,24 @@
+package br
+
+import (
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/validation"
+)
+
+// invoiceValidator adds validation checks to invoices which are relevant
+// for the region.
+type invoiceValidator struct {
```
samlown, line 10: Is this validator even required? Certainly its not doing anything useful at the moment.

samlown (approved):
Looks good, but I'd strongly suggest removing the invoice validation file as it doesn't do anything and is unlikely to be needed. Thinking about how to do the addon in BR could be tougher (`br-nfs-v1` maybe?)

Also, we want to avoid reducing test coverage ;-)

regimes/br/invoices.go
cavalle: I guess you referred to the invoiceValidator struct but not the validations themselves. I copied the code from the US regime so I've simplied both in [ce4698c](https://github.com/invopop/gobl/pull/396/commits/ce4698cf9fdcf004e14210e12f8c5552200acc48).

As for the coverage, I reached ~95% with [2103b22](https://github.com/invopop/gobl/pull/396/commits/2103b22b629aeb7bf47377d7bbb9019ef5abdea2). Not sure why codecov still complains.

cavalle: I needed to merge `main` to the branch. Codecov should be fine now.

PR #397, opened 2024-10-18
Add extension for EsigibilitaIVA to it-sdi addon

cavalle:


addons/it/sdi/extensions.go
```diff
@@ -15,6 +15,7 @@ const (
 	ExtKeyExempt       cbc.Key = "it-sdi-exempt"
 	ExtKeyRetained     cbc.Key = "it-sdi-retained"
 	ExtKeyPaymentMeans cbc.Key = "it-sdi-payment-means"
+	ExtKeyVATCollect   cbc.Key = "it-sdi-vat-collect"
```
samlown, line 18: I'm not super-convinced on "collect" as its a verb... but can't think of anything more concise...

samlown (approved):

addons/it/sdi/extensions.go
samlown: Just checking this in ChatGTP, the response is:
> “Esigibilità dell’IVA” translates to “VAT enforceability” or “VAT liability.” It refers to the moment when VAT becomes due or payable to the tax authorities.

cavalle: Addressed on 7b87944

PR #399, opened 2024-10-21
Set proper regime when alternative code is given

cavalle:
## Describe your changes

Sets the proper tax regime code (_e.g._ `EL`) when the alternative code is given (_e.g._ `GR`)


samlown (approved):
Nice! That's what I was expecting 👍

PR #400, opened 2024-10-22
Adding TIN algorithms

Menendez6:
## Describe your changes

I added TINs algorithms in identities for Austria, France, Great Britain and Poland and I have created a first version of the regime in Ireland.

- [ ] I have performed a self-review of my code.
- [ ] I have added thorough tests with at **least** 90% code coverage.
- [ ] I've modified or created example GOBL documents the show my changes in use, if appropriate.
- [ ] When adding or modifying a tax regime or addon, I've added links to the source of the changes either structured or in the comments.
- [ ] I've run `go generate .` to ensure that Schemas and Regime data are up to date.

### DIFF
--- a/regimes/at/identitites.go
+++ b/regimes/at/identitites.go
@@ -1,0 +1,80 @@
+package at
+
+import (
+	"errors"
+	"regexp"
+	"strconv"
+
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/validation"
+)
+
+const (
+	// IdentityKeyTaxNumber represents the Austrian tax number (Steuernummer) issued to
+	// people that can be included on invoices inside Austria. For international
+	// sales, the registered VAT number (Umsatzsteueridentifikationsnummer) should
+	// be used instead.
+	IdentityKeyTaxNumber cbc.Key = "at-tax-number"
+)
+
+var badCharsRegexPattern = regexp.MustCompile(`[^\d]`)
+
+var identityKeyDefinitions = []*cbc.KeyDefinition{
+	{
+		Key: IdentityKeyTaxNumber,
+		Name: i18n.String{
+			i18n.EN: "Tax Number",
+			i18n.DE: "Steuernummer",
+		},
+	},
+}
+
+func normalizeTaxNumber(id *org.Identity) {
+	if id == nil || id.Key != IdentityKeyTaxNumber {
+		return
+	}
+	code := id.Code.String()
+	code = badCharsRegexPattern.ReplaceAllString(code, "")
+	id.Code = cbc.Code(code)
+}
+
+func validateTaxNumber(id *org.Identity) error {
+	if id == nil || id.Key != IdentityKeyTaxNumber {
+		return nil
+	}
+
+	return validation.ValidateStruct(id,
+		validation.Field(&id.Code, validation.By(validateTaxIdCode)),
+	)
+}
+
+// validateAustrianTaxIdCode validates the normalized tax ID code.
+func validateTaxIdCode(value interface{}) error {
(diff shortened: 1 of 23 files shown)


regimes/at/at.go
```diff
@@ -50,7 +52,10 @@ func Validate(doc any) error {
 		return validateInvoice(obj)
 	case *tax.Identity:
 		return validateTaxIdentity(obj)
+	case *org.Identity:
+		return validateTaxNumber(obj)
```
samlown, line 56: This should be:
```suggestion
		return validateIdentity(obj)
```

regimes/at/identitites.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/validation"
+)
+
+const (
+	// IdentityKeyTaxNumber represents the Austrian tax number (Steuernummer) issued to
+	// people that can be included on invoices inside Austria. For international
+	// sales, the registered VAT number (Umsatzsteueridentifikationsnummer) should
+	// be used instead.
+	IdentityKeyTaxNumber cbc.Key = "at-tax-number"
```
samlown, line 19: I'm thinking that these should be Identity types also, in this case ChatGTP suggests `STNR`. The same could also be applied to Germany (but we don't want to do that yet!)

regimes/at/identitites.go
```diff
@@ -0,0 +1,82 @@
...
+	{
+		Key: IdentityKeyTaxNumber,
+		Name: i18n.String{
+			i18n.EN: "Tax Number",
+			i18n.DE: "Steuernummer",
+		},
+	},
+}
+
+func normalizeTaxNumber(id *org.Identity) {
```
samlown, line 36: There needs to be a generic normalize method. Naming also needs to be consistent, so this should be `normalizeIdentity`.

regimes/gb/identities.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if !matched {
+		return errors.New("invalid UTR format")
+	}
+
+	return nil
+}
+
+// validateNino validates the normalized National Insurance Number (NINO).
+func validateNino(value interface{}) error {
```
samlown, line 102: I'd stick to the `validateIdentity` convention and check the type.

regimes/gb/identities.go
```diff
@@ -0,0 +1,157 @@
...
+		id.Code = cbc.Code(code)
+	}
+}
+
+func validateIdentity(id *org.Identity) error {
+	if id == nil {
+		return nil
+	}
+
+	if id.Type == IdentityTypeNINO {
```
samlown, line 66: Switches are always clearer in these circumstances.

regimes/pl/pl.go
```diff
@@ -72,6 +74,8 @@ func Validate(doc interface{}) error {
 		return validateTaxIdentity(obj)
 	case *bill.Invoice:
 		return validateInvoice(obj)
+	case *org.Identity:
+		return validateTaxNumber(obj)
```
samlown, line 78:
```suggestion
		return validateIdentity(obj)
```

samlown (changes requested):
Lot's going on here... we need consistency on naming, and I'd also try to standardise on using type codes instead of keys for local identities.

Also, patch test coverage *IS NOT* 90%!

regimes/pl/identities.go
```diff
@@ -0,0 +1,69 @@
...
+	IdentityTypePESEL cbc.Code = "PESEL"
+)
+
+// Reference: https://en.wikipedia.org/wiki/PESEL
+
+var identityTypeDefinitions = []*cbc.ValueDefinition{
+	{
+		Value: IdentityTypePESEL.String(),
+		Name: i18n.String{
+			i18n.EN: "Tax Number",
```
samlown, line 25: I think this name needs to be a bit more descriptive.

PR #402, opened 2024-10-25
Add initial Brazil NFSe addon

cavalle:

### DIFF
--- a/addons/br/nfse/extensions.go
+++ b/addons/br/nfse/extensions.go
@@ -1,0 +1,32 @@
+package nfse
+
+import (
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/pkg/here"
+)
+
+// Brazilian extension keys required to issue NFS-e documents.
+const (
+	ExtKeyService = "br-nfse-service"
+)
+
+var extensions = []*cbc.KeyDefinition{
+	{
+		Key: ExtKeyService,
+		Name: i18n.String{
+			i18n.EN: "Service Code",
+			i18n.PT: "Código Item Lista Serviço",
+		},
+		Desc: i18n.String{
+			i18n.EN: here.Doc(`
+				The service code as defined by the municipality. Typically, one of the codes listed
+				in the Lei Complementar 116/2003, but municipalities can make their own changes.
+
+				For further details on the list of possible codes, see:
+
+				* https://www.planalto.gov.br/ccivil_03/leis/lcp/lcp116.htm
+			`),
+		},
+	},
+}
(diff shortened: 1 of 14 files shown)


addons/br/nfse/extensions.go
```diff
@@ -0,0 +1,32 @@
...
+		Desc: i18n.String{
+			i18n.EN: here.Doc(`
+				The service code as defined by the municipality. Typically, one of the codes listed
+				in the Lei Complementar 116/2003, but municipalities can make their own changes.
+
+				For further details on the list of possible codes, see:
+
+				* https://www.planalto.gov.br/ccivil_03/leis/lcp/lcp116.htm
+			`),
+		},
```
samlown, line 30: Is is possible to add a `Pattern` flag perhaps to validate the codes?

addons/br/nfse/line.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func validateLine(value any) error {
+	line, _ := value.(*bill.Line)
+	if line == nil {
+		return nil
+	}
+
+	return validation.Validate(line,
+		bill.RequireLineTaxCategory(br.TaxCategoryISS),
```
samlown, line 16: Does this also need to be repeated for Discount and Charge and lines?

addons/br/nfse/item_test.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+		{
+			name: "missing extension",
+			item: &org.Item{
+				Ext: tax.Extensions{
+					"random": "12345678",
+				},
+			},
+			err: "ext: (br-nfse-service: required.).",
+		},
```
samlown, line 46: You could test the nil case here like this:
```suggestion
		},
		{
		  name: "nil",
		  item: nil,
		},
```

samlown (approved):
Nice. Some observations. Do the Charges and Discounts also need to be covered?

addons/br/nfse/extensions.go
cavalle: I chose not to validate anything since every municipality can define their own list of codes without restrictions on the format. The list in LEI116 is commonly used, apparently, but there's no guarantee those will be the codes in the municipality.

addons/br/nfse/line.go
cavalle: Actually, I don't think it makes sense to support either Charges or Discounts: it doesn't seem possible to map them to an NFSe. So, I've added validations to ensure those aren't present in [177eb69](https://github.com/invopop/gobl/pull/402/commits/177eb697a8a4bf0fecd5edf6bc1cab717d7fa216)

cavalle:
@samlown Thanks for your feedback! I think I addressed all your points and made an additional change (derived from your feedback as well) that let me get to 98% coverage.

PR #410, opened 2024-11-04
Veri*Factu Addon

apardods:
## Describe your changes

Add-on for the Ver*Factu Regulation. Includes:

- Extensions: `es-verifactu-doc-type` for document type and `es-verifactu-tax-classification` to cover tax exemptions and overall classification. They correspond to L2 and L9/10 from the [BOE order](https://www.boe.es/diario_boe/txt.php?id=BOE-A-2024-22138), at the very bottom.
- Invoice Validation: 
   - Correction invoices must include Preceding, which must include the issue date. 
   - Invoice Tax field must include the `es-verifactu-doc-type` extension.
   - Line taxes must include the `es-verifactu-tax-classification` extension.
   - There must be a note with key `general` that, like with TicketBAI, acts as the invoice description.
   - If there is a customer with a Spanish taxID, require its code
- Scenarios: automatically apply the `es-verifactu-doc-type` extension based on invoice information.
- Tags: added a Substitution tag for `Factura emitida en sustitución de facturas simplificadas facturadas y declaradas.`, to cover the case when a client that has received a simplified invoice then asks for a full one.

- [ ] ~~I have added thorough tests with at **least** 90% code coverage.~~ Coverage tops out at 87% because of invoice nil checks. Could add tests although they would be somewhat redundant.


apardods (commented):
Looks great! I think it makes sense to only cover the most basic cases, and to leave things like reverse charges to the user, as using scenarios was risky.

samlown (approved):
🔝 🎩

PR #416, opened 2024-11-06
Moving examples from regimes to single folder

samlown:
* Moving all the examples previously in the regimes, to a single `/examples` directory.


cavalle (approved):
LGTM!

PR #421, opened 2024-11-12
Arab emirates regime

richifernandez:

- [X] I've modified or created example GOBL documents the show my changes in use, if appropriate.
- [X] When adding or modifying a tax regime or addon, I've added links to the source of the changes either structured or in the comments.
- [X] I've run `go generate .` to ensure that Schemas and Regime data are up to date.


regimes/ae/tax_identity.go
```diff
@@ -0,0 +1,39 @@
...
+func validateTRNCode(value interface{}) error {
+	code, ok := value.(cbc.Code)
+	if !ok || code == "" {
+		return nil
+	}
+	val := code.String()
+
+	// Check if TRN matches the 15-digit pattern
+	if !trnRegex.MatchString(val) {
+		return errors.New("invalid format: TRN must be a 15-digit number")
```
samlown, line 35: I'd suggest simplifying the error message here to just:
```suggestion
		return errors.New("must be a 15-digit number")
```

CHANGELOG.md
```diff
@@ -599,3 +599,7 @@ New number formatting support! Expect some possible breaking SDK changes with th
 ## [v0.70.1] - 2024-03-25
 
 - Last version before CHANGELOG.md.
+
+### Added
+
+- ae: added UAE regime
```
samlown, line 605: This line should be at the top of the document, in the `[Unreleased]` section.

examples/ae/invoice-ae-ae-stnr.yaml
```diff
@@ -0,0 +1,48 @@
+$schema: "https://gobl.org/draft-0/bill/invoice"
```
samlown, line 1: I can't see the output from these files, did you run `go generate ./...`?

samlown (changes requested):
A few little changes required, specifically to ensure the example output are generated, but otherwise looking good! 👍

examples/ae/invoice-ae-ae-stnr.yaml
richifernandez: I have implemented the JSON output for the examples. However, I had to remove the Arabic version of the notes in scenario.go because it wasn't generating the Arabic characters correctly.

Below is how the notes appeared originally:
gobl-ae-scenarios go


And here is how the incorrect characters were generated:
gobl-ae-example-invoice go

samlown (approved):
👍  Looking good! Nice one. Strange merge thing with CHANGELOG, but otherwise okay.

PR #451, opened 2025-01-19
Singapore

alvarolivie:
## Describe your changes

Implemented the regime for Singapore following the documentation provided by [IRAS](https://www.iras.gov.sg/taxes/goods-services-tax-(gst))

- [ ] I've modified or created example GOBL documents the show my changes in use, if appropriate.

## Considerations

I am not sure if receipts need to be added as an invoice. IRAS states that they are a valid form of invoicing in certain situations. I followed the format used in Portugal as they have a similar type of invoice.

I did not divide the different type of GST registration numbers as different Identities as they are all GST reg. numbers even if they are also used elsewhere.

### DIFF
--- a/regimes/sg/README.md
+++ b/regimes/sg/README.md
@@ -1,3 +1,5 @@
 # 🇸🇬 GOBL Singapore Tax Regime
 
 
+
+
--- a/regimes/sg/README.md
+++ b/regimes/sg/README.md
@@ -20,6 +20,8 @@ Singapore offers a simple GST model with a standard rate along with a few except
 
 ## Invoicing methods
 
+There are three main methods for invoicing which will be described below. Other methods like credit notes and reverse billing have to follow the structure of a normal Tax Invoice.
+
 ### Tax Invoice
 
 This invoice, reference in GOBL by the use of the tax tag "standard" represents a basic Invoice. This Invoice has to meet certain requirements:
--- a/regimes/sg/README.md
+++ b/regimes/sg/README.md
@@ -1,0 +1,3 @@
+# 🇸🇬 GOBL Singapore Tax Regime
+
+
--- a/regimes/sg/README.md
+++ b/regimes/sg/README.md
@@ -58,4 +58,12 @@ This type of invoice can be issued to a non-GST registered costumer. This invoic
 
 ### GST Registration Number
 
-There are multiple possiblities when it comes to GST reg nums. They can be a Unique Entity Number (UEN) which refers to business, they can be a National Registration Identity Card (NRIC) number or a Foreign Identification Number which refer to people, and they can be a proper GST reg num emmited by IRAS. Each code has its own validation rules along which have been implemented
+There are multiple possiblities when it comes to GST reg nums. They can be a Unique Entity Number (UEN) which refers to business, they can be a National Registration Identity Card (NRIC) number or a Foreign Identification Number which refer to people, and they can be a unique GST reg num emmited by IRAS. Each code has its own validation rules along which have been implemented. Each type of code is still a valid GST registration number and must be included in all invoices.
+
+### References
+
+[GST General Guide for Businesses](https://www.iras.gov.sg/media/docs/default-source/e-tax/etaxguide_gst_gst-general-guide-for-businesses(1).pdf?sfvrsn=8a66716d_97)
+
+[GST Rates](https://www.iras.gov.sg/taxes/goods-services-tax-(gst)/basics-of-gst/current-gst-rates)
+
+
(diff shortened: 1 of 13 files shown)


regimes/sg/README.md
```diff
@@ -0,0 +1,75 @@
...
+6. A description sufficient to identify the goods or services supplied and the type of supply.
+7. For each description of goods or services supplied, the quantity of goods or the extent of services, and the amount payable, excluding GST.
+8. Any cash discount offered.
+9. The total amount payable (excluding GST), the GST rate and the total amount of GST chargeable.
+10. The total amount payable (including the total amount of GSTchargeable).
+11. A breakdown of exempt, zero-rated or other supplies, stating separatelythe gross total amount payable in respect of each type of supply.
+
+### Simplified Tax Invoice
+
+This invoice is referenced by the tax tag "simplified". This invoice can only be used when the total amount (inclusive of GST) is less than $1000. This invoice has less requirements:
```
samlown, line 45: It'd be useful to add the currency code of $1000 here.

regimes/sg/README.md
```diff
@@ -0,0 +1,75 @@
...
+6. The word “Price Payable includes GST”.
+
+### Reciept
+
+This type of invoice can be issued to a non-GST registered costumer. This invoice requires the following:
+
+1. Suplier name, address and GST registration number;
+2. The date of issue of the invoice.
+3. The total amount payable including tax.
+4. The word “Price Payable includes GST”.
```
samlown, line 61: Based on this description, it feels like the receipt is probably not needed in GOBL, the "simplified" version will cover most cases.

regimes/sg/sg.go
```diff
@@ -0,0 +1,65 @@
...
+			},
+		},
+		Validator:  Validate,
+		Normalizer: Normalize,
+		Categories: taxCategories,
+	}
+}
+
+// Validate checks the document type and determines if it can be validated.
+func Validate(doc interface{}) error {
```
samlown, line 49: tiny observation:
```suggestion
func Validate(doc any) error {
```

regimes/sg/tax_categories.go
```diff
@@ -0,0 +1,72 @@
...
+				Key: tax.RateStandard,
+				Name: i18n.String{
+					i18n.EN: "Standard rate",
+				},
+				Description: i18n.String{
+					i18n.EN: "For the majority of sales of goods and services: it applies to all products or services for which no other rate is expressly provided.",
+				},
+				Values: []*tax.RateValueDef{
+					{
+						Since:   cal.NewDate(2024, 1, 1),
```
samlown, line 55: This is very recent. It wouldn't hurt to include the last standard rate if available for the sake of backwards compatibility.

regimes/sg/tax_identity.go
```diff
@@ -0,0 +1,56 @@
...
+func validateTaxIdentity(tID *tax.Identity) error {
+	return validation.ValidateStruct(tID,
+		validation.Field(&tID.Code,
+			validation.By(validateTaxCode),
+			validation.Skip,
+		),
+	)
+}
+
+func validateTaxCode(value interface{}) error {
```
samlown, line 37: Is there no checksum validation in SG for any of the codes?

regimes/sg/tax_identity.go
```diff
@@ -0,0 +1,56 @@
...
+// Reference: https://mytax.iras.gov.sg/ESVWeb/default.aspx?target=GSTListingSearch
+// All these tax codes are valid options for a GST registration number.
+
+var (
+	taxCodeRegexps = []*regexp.Regexp{
+		regexp.MustCompile(`^(19[0-9]{2}|20[0-9]{2})\d{5}[A-Z]$`), // UEN (ROC)
+		regexp.MustCompile(`^\d{9}[A-Z]$`),                        // UEN (ROB)
+		regexp.MustCompile(`^[TS]\d{2}[A-Z]\w{1}\d{4}[A-Z]$`),     // UEN (Others)
+		regexp.MustCompile(`^[STFGM]\d{7}[A-Z]$`),                 // NIRC/FIN
+		regexp.MustCompile(`^\w{2}\d{7}[A-Z]$`),                   // GST
```
samlown, line 22: With so many to choose from, I get the impression that maybe some companies choose to use multiple numbers at the same time on their invoices, suggesting that only one should be used for the main tax identity.

It could be that the GST makes the most sense in the tax identity, and everything else should be moved to the `org.Identity`. An additional (not obvious) option here is to switch to using the `identities` array instead of the `tax.Identity`.

samlown (commented):
This is a great first attempt! Nice one! I'd question if the concept of "receipt" should be there, it feels like something that would be used with a payment device rather than for invoicing.

I'm also unsure about the identity code handling given so many options.

PR #457, opened 2025-02-11
Payment receipts and advances in Portugal

cavalle:
- Adds support for payment advances and receipts in PT
- Changes the default unit to `one` in PT
- Fixes a panic when calculating a `bill.Receipt` without lines
- Changes `bill.Receipt` to allow empty codes while the invoice isn't signed (like in `bill.Invoice`)
- Calculates line indexes automatically in `bill.Receipt`

## Pre-Review Checklist


regimes/pt/README.md
```diff
@@ -15,7 +15,7 @@ Find example PT GOBL files in the [`examples`](../../examples/pt) (uncalculated
 
 ### `InvoiceType` (Tipo de documento)
 
-SAF-T's `InvoiceType` (Tipo de documento) specifies the type of a Portuguese tax document. The following table lists all the supported invoice types and how GOBL will map them with a combination of invoice type and tax tags:
+SAF-T's `InvoiceType` (Tipo de documento) specifies the type of a sales invoice. In GOBL, this type can be set using the `pt-saft-invoice-type` extension in the tax section. GOBL will set the extension for you based on the type and the tax tags you set in your invoice. The table below shows how this mapping is done:
```
samlown, line 18: Could all of this README stuff be added to the Addon description and extensions like for GR now? This way it'll get published automatically to the documentation site. I understand there are also specific for the addon, and probably shouldn't be in the Regime.

samlown (changes requested):
Code looks great! I do think we should start moving all the documentation to the structure approach though. My objective is to remove README.md files completely, as everything should be defined inside the Go structures.

regimes/pt/README.md
cavalle: Cool, I'll move PT taking GR as a reference

cavalle: Addressed in [43c18e3](https://github.com/invopop/gobl/pull/457/commits/43c18e37a7df7c17d49d32e07bf21e9d05f7d395)

addons/pt/saft/extensions.go
```diff
@@ -166,6 +245,73 @@ var extensions = []*cbc.Definition{
...
 			i18n.PT: "Código do motivo de isenção de imposto",
 		},
+		Desc: i18n.String{
+			i18n.EN: here.Doc(`
+				AT's ~TaxExemptionCode~ (Código do motivo de isenção de imposto) is a code that
+				specifies the reason the VAT tax is exempt in a Portuguese invoice. When the ~exempt~ tag
+				is used, one of the following must be defined in the ~ext~ map's ~pt-exemption-code~
+				property:
+
+				| Code  | Description                                                                                              |
```
samlown, line 255: So listing the table again is probably redundant all of these will be presented in the output, indeed, they already are: https://docs.gobl.org/addons/pt-saft-v1#pt-saft-exemption

The exception to this rule is when there is a specific mapping using special keys, and the link is not obvious.

addons/pt/saft/extensions.go
```diff
@@ -420,6 +603,73 @@ var extensions = []*cbc.Definition{
...
 			i18n.PT: "Meio de Pagamento",
 		},
+		Desc: i18n.String{
+			i18n.EN: here.Doc(`
+				The SAF-T's ~PaymentMechanism~ (Meios de pagamento) field specifies the payment means in a
+				sales invoice or payment. GOBL provides the ~pt-saft-payment-means~ extension to set this
+				value in your ~bill.Invoice~ advances or in you ~bill.Receipt~ method. GOBL maps certain
+				payment mean keys automatically to this extension:
+
+				| Code | Name                                               | GOBL Payment Mean                                     |
```
samlown, line 613: So this table for example probably does make sense as there is a specific mapping from keys to types... is it worth considering reversing the order? Ideally you'd want to know what Payment Means `key` to use and the resulting code.

samlown (commented):
I think we can be a bit more efficient perhaps with the tables, but otherwise, this looks great! Juan will be happy too :-)

samlown (approved):

PR #460, opened 2025-02-17
new addon for italian scontrino

alvarolivie:
The adecf addon is currently based on information from the [AdE](https://www.agenziaentrate.gov.it/portale/home).
As no official Italian framework for uploading this type of document is yet in place, this addon must be treated as a work in progress. We cannot be sure that the format that the Italian government require will not change.

Cannot check for nil supplier or nil tax as GOBL will reject the invoice before it reaches this addon's validation. 

## Pre-Review Checklist


addons/it/adecf/adecf.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func init() {
+	tax.RegisterAddonDef(newAddon())
+}
+
+// This validation follows the rules of the Italian Agenzia delle Entrate
+// This addon will then be used to create documents using the following services
+// https://www.agenziaentrate.gov.it/portale/servizi/servizitrasversali/altri/cassetto-fiscale
+// https://www.agenziaentrate.gov.it/portale/schede/comunicazioni/fatture-e-corrispettivi
```
samlown, line 26: These could be added as sources in the AddonDef

addons/it/adecf/adecf.go
```diff
@@ -0,0 +1,55 @@
...
+			i18n.EN: "Italy AdE adecf v1.x",
+		},
+		Extensions: extensions,
+		Validator:  validate,
+		Normalizer: normalize,
+	}
+}
+
+func normalize(doc any) {
+	// nothing to normalize yet
```
samlown, line 40: not needed.

addons/it/adecf/adecf.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/gobl/tax"
+)
+
+// Key to identify the AdE adecf addon
+const (
+	// V1 for AdE format
+	V1 cbc.Key = "it-adecf-v1"
```
samlown, line 16: The `adecf` name looks highly unreadable in this context. Would `it-ticket-v1` work?

addons/it/adecf/invoices.go
```diff
@@ -0,0 +1,62 @@
...
+	return validation.ValidateStruct(inv,
+		validation.Field(&inv.Tax,
+			validation.Required,
+			validation.By(validateInvoiceTax),
+			validation.Skip,
+		),
+		validation.Field(&inv.Supplier,
+			validation.By(validateInvoiceSupplier),
+			validation.Skip,
+		),
```
samlown, line 20: Do we need to check that the customer is empty or correct?

addons/it/adecf/invoices_test.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func TestInvoiceTax(t *testing.T) {
+	t.Run("invalid PricesInclude", func(t *testing.T) {
+		inv := exampleStandardInvoice(t)
+		inv.Tax.PricesInclude = "invalid"
+		require.NoError(t, inv.Calculate())
+		err := inv.Validate()
+		require.EqualError(t, err, "tax: (prices_include: must be a valid value.).")
+	})
```
samlown, line 173: This test feels a bit out of scope, and should already be validated elsewhere.

addons/it/adecf/invoices.go
```diff
@@ -0,0 +1,62 @@
...
+func validateInvoiceTax(value interface{}) error {
+	t, ok := value.(*bill.Tax)
+	if !ok || t == nil {
+		return nil
+	}
+
+	return validation.ValidateStruct(t,
+		validation.Field(&t.PricesInclude,
+			validation.Required,
+			validation.In(tax.CategoryVAT),
```
samlown, line 57: Good to validate. Probably makes sense to normalize the GOBL document and add the PricesInclude automatically with VAT.

addons/it/adecf/tax.go
```diff
@@ -0,0 +1,69 @@
...
+	num.MakePercentage(22, 2),
+}
+
+func validateTaxCombo(val any) error {
+	c, ok := val.(*tax.Combo)
+	if !ok || c == nil {
+		return nil
+	}
+
+	if c.Category == tax.CategoryVAT {
```
samlown, line 37: I'd usually reverse these types of `if` to avoid embedding. Checking you don't have VAT and exiting is probably easier to read.

```
if c.Category != tax.CategoryVAT {
  return nil
}
```

addons/it/adecf/tax.go
```diff
@@ -0,0 +1,69 @@
...
+	if !ok || p == nil {
+		return nil
+	}
+
+	for _, vp := range validPercentages {
+		if p.Compare(vp) == 0 {
+			return nil
+		}
+	}
+	return validation.NewError("validation_percentage_percentage", "Invalid percentage")
```
samlown, line 68: This error code is a bit strange (and at very least repetitive). Probably best just to return a simple error:
```suggestion
	return errors.New("must be a valid value")
```

data/schemas/currency/code.json
```diff
@@ -38,12 +38,12 @@
           "title": "Macanese Pataca"
         },
         {
-          "const": "HTG",
-          "title": "Haitian Gourde"
+          "const": "BWP",
```
samlown, line 41: I'd try and avoid these types of random generate changes if possible. Would be good to understand why this is happening.

samlown (changes requested):
Looking great! Almost there. A few minor suggestions. For sure we should rename.

samlown (approved):
All looking good 👍  nice!

PR #496, opened 2025-03-20
added validation for non latin chars

alvarolivie:

## Pre-Review Checklist


CHANGELOG.md
```diff
@@ -8,12 +8,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
...
 ### Added
 
+- `it-sdi-v1`: validation for non latin and latin-1 characters in item names.
 - `pt-saft-v1`: support for stock movements (from `bill.Delivery`) and work documents (from `bill.Invoice` and `bill.Order`)
 
 ## [v0.212.1] - 2025-03-11
 
 ### Changed
 
+- `bill`: line rounding will now check for nil item and nil item price.
```
samlown, line 18: Is this in the right place?

bill/line_calculate.go
```diff
@@ -275,18 +275,17 @@ func roundLines(lines []*Line, cur currency.Code) {
 // round will automatically apply the correct rounding rules.
 func (l *Line) round(cur currency.Code) {
 	e := cur.Def().Subunits
-	if l.Item.Price != nil {
+	if l.Item != nil && l.Item.Price != nil {
 		e = l.Item.Price.Exp()
+		if l.Total != nil {
+			total := l.Total.RescaleDown(e)
+			l.Total = &total
+		}
```
samlown, line 283: I'm not sure this is the best approach. I'd suggest leaving the `if l.Total != nil` condition where it was, and either check the for the presence of the `l.Item.Price` or return immediately if there is no line item or price.

bill/line_calculate_test.go
```diff
@@ -571,3 +571,109 @@ func TestLineCalculate(t *testing.T) {
 		assert.Equal(t, "39.77", lines[0].Total.String())
 	})
 }
+
+func TestRoundLines(t *testing.T) {
```
samlown, line 575: This test feels a bit strange as it looks like a feature, when in actual fact its avoiding a bug. The underlying issue is that the lines were not correctly normalised and there are still totals and sums, despite the lack of an item. I think a better approach will be to abort the rounding early if there is nothing useful to round. To really fix the bug, normalization should completely remove the sum and total if there is not item nor price.

samlown (changes requested):
SDI invoice stuff looks good. The changes for the line calculations feel strange. I'd suggest trying to fix the underlying bug, normalization, and ensure the rounding method will simply abort if it has invalid data.

bill/line_calculate.go
alvarolivie: I agree, the results this function provided did not really make sense.

bill/line_calculate_test.go
alvarolivie: TestLineCalculate already tests most of (l *line) round(). I think we should remove these tests and simply add to the missing item test so that we check that round returns early if item or item.price is nil

CHANGELOG.md
alvarolivie: No, clearly not.

samlown (approved):
🔝

PR #516, opened 2025-04-24
Refactor: adding 'Retained' field to tax totals

samlown:
- New `Retained` field in tax totals.
- New `RetainedTax` field in bill totals.
- Calculations adjust to separate taxes between indirect and withheld/retained taxes.
- Removed the redundant `PreciseSum()` method from tax totals, which is no longer used.

## Pre-Review Checklist


bill/invoice_test.go
```diff
@@ -685,10 +685,9 @@ func TestRemoveIncludedTaxDeep(t *testing.T) {
 		t.Logf("TOTALS: %v", string(data))
 	*/
 
+	assert.Equal(t, "17830.19", i.Totals.Total.String())
 	assert.Equal(t, "17830.20", i2.Totals.Total.String())
-	// assert.Equal(t, i.Totals.Total.String(), i2.Totals.Total.String())
-	assert.Equal(t, "17830.20", i2.Totals.Total.String())
-	assert.Equal(t, "-0.02", i2.Totals.Rounding.String())
+	assert.Equal(t, "-0.01", i2.Totals.Rounding.String())
```
samlown, line 690: Not sure why this is happening, but it is an improvement.

cavalle (approved):
I think the segregation of taxes and retained taxes makes sense and it's matches what we usually see in e-invoice or e-reporting formats. My only comments would be:

1. I'd make sure we understand why the calculations/rounding before and after have changed and we're fine with them.
2. The `totals_with_taxes` and `tax` totals now don't incude retained taxes. I guess that's probably ok as it's very likely no converter is using those values assuming the retentions are accounted for in those values. But, strictly speaking, this is a backward incompatible change which could potentially break something. Just wanted to highlight that potential risk (which I'm sure was already considered)

bill/invoice_test.go
samlown: It shouldn't... the PreciseSum method wasn't actually doing anything since the last set of changes.

PR #517, opened 2025-04-25
Fix addons

alvarolivie:
This PR looks to add all the missing rules and validations schematrons impose on EN16931 formats. It combines previous PRs (#511 and #509 ) which have now been closed.

This PR:
- Adds EN16931 rules which where not being passed naturally by GOBL during testing and needed to be enforced
- Adds identity validation and normalization to FR (SIREN/SIRET)
- Minor addition in SDI regarding postbox
- Adds al the missing Xrechnung rules

## Pre-Review Checklist


addons/eu/en16931/bill.go
```diff
@@ -104,3 +112,61 @@ func validateBillInvoiceTax(value any) error {
...
+}
+
+func validateBillLineCharge(value any) error {
+	// BR-44
+	charge, ok := value.(*bill.LineCharge)
+	if !ok || charge == nil {
+		return nil
+	}
+	if charge.Reason == "" && (charge.Ext == nil || charge.Ext[untdid.ExtKeyCharge] == "") {
+		return validation.NewError("reason", "either a reason or a charge type extension is required")
```
samlown: It could be useful to include the `BR-44` instead of `reason` in the New Error call.

samlown (approved):
Looks great. Potential consideration for the NewError codes which could use the same business rule codes. I don't think we should merge this however until we've had chance to investigate any potential issues with users and the new validation rules.

addons/eu/en16931/bill.go
alvarolivie: Done!

addons/eu/en16931/bill.go
```diff
@@ -104,3 +112,61 @@ func validateBillInvoiceTax(value any) error {
...
+	)
+}
+
+func validateBillLineCharge(value any) error {
+	// BR-44
+	charge, ok := value.(*bill.LineCharge)
+	if !ok || charge == nil {
+		return nil
+	}
+	if charge.Reason == "" && (charge.Ext == nil || charge.Ext[untdid.ExtKeyCharge] == "") {
```
samlown: I'd use `!charge.Ext.Has(untdid.ExtKeyCharge)` here for readability..

addons/eu/en16931/bill.go
```diff
@@ -104,3 +112,61 @@ func validateBillInvoiceTax(value any) error {
...
+	return nil
+}
+
+func validateBillLineDiscount(value any) error {
+	// BR-41
+	discount, ok := value.(*bill.LineDiscount)
+	if !ok || discount == nil {
+		return nil
+	}
+	if discount.Reason == "" && (discount.Ext == nil || discount.Ext[untdid.ExtKeyAllowance] == "") {
```
samlown: Same here:
```suggestion
	if discount.Reason == "" && !discount.Ext.Has(untdid.ExtKeyAllowance) {
```

addons/eu/en16931/pay.go
```diff
@@ -43,3 +44,31 @@ func validatePayInstructions(instr *pay.Instructions) error {
 		),
 	)
 }
+
+func validateBillPayment(payment *bill.PaymentDetails) error {
```
samlown: I think this method should be in the `bill.go` file, not `pay.go`.

addons/eu/en16931/pay.go
```diff
@@ -43,3 +44,31 @@ func validatePayInstructions(instr *pay.Instructions) error {
...
+	return validation.ValidateStruct(payment,
+		validation.Field(&payment.Terms,
+			validation.Required,
+			validation.By(validateBillPaymentTerms),
+			validation.Skip,
+		),
+	)
+}
+
+func validateBillPaymentTerms(value any) error {
```
samlown: This should probably be called directly from the main validation method as it looks quite generic.

addons/eu/en16931/bill.go
alvarolivie: Makes sense!

addons/eu/en16931/pay.go
alvarolivie: Yes, I think it's more clear.

addons/eu/en16931/pay.go
alvarolivie: I was not sure if another object used pay.Terms. Done!

PR #518, opened 2025-04-27
Add Sweden regime

BoscoDomingo:

## Pre-Review Checklist

- [X] When adding or modifying a tax regime or addon, I've added links to the source of the changes, either structured or in the comments.
- [X] I've run `go generate .` to ensure that the Schemas and Regime data are up to date.


regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+		validation.Field(&inv.Customer,
+			validation.Required,
+			validation.By(v.validateCommon),
+			validation.By(v.validateCustomer),
+			validation.Skip,
+		),
+	)
+}
+
+func (v *invoiceValidator) validateCommon(value any) error {
```
samlown, line 39: Usually best to identify the underlying object type: `validateOrgParty`

regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+				tID, ok := value.(*tax.Identity)
+				if !ok || tID == nil {
+					return nil
+				}
+				return validateTaxIdentity(tID)
+			}),
+		),
+		validation.Field(&party.Identities,
+			validation.Required,
+			org.RequireIdentityKey(ValidOrgIdentityKeys...),
```
samlown, line 78: Is this true? Usually only the VAT ID is a requirement, everything else is optional.

regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+	party, ok := value.(*org.Party)
+	if !ok || party == nil {
+		return nil
+	}
+
+	return validation.ValidateStruct(party,
+		validation.Field(&party.Identities,
+			validation.When(
+				party.TaxID != nil,
+				org.RequireIdentityKey(ValidOrgIdentityKeys...),
```
samlown, line 95: We're missing the country check. Foreign companies and not going to have Swedish IDs. I also wouldn't expect this to be a country level requirement.

internal/luhn.go
```diff
@@ -0,0 +1,38 @@
...
+var nonNumberRegExp = regexp.MustCompile(`[^\d]+`)
+
+// ValidateLuhn checks if a string of digits passes the Luhn algorithm validation.
+//
+// `number` must contain only numeric characters.
+//
+// [Source]
+//
+// [Source]: https://github.com/luhnmod10/go
+func ValidateLuhn(number string) bool {
```
samlown, line 16: The `/regimes/common/luhn.go` already implements this.

regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (v *invoiceValidator) validateCommon(value any) error {
+	party, ok := value.(*org.Party)
+	if !ok || party == nil {
+		return nil
+	}
+
+	return validation.ValidateStruct(party,
+		validation.Field(&party.Name,
+			validation.Required,
```
samlown, line 47: missing `validation.Skip`

regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+		),
+		validation.Field(&party.Addresses,
+			validation.Required,
```
samlown, line 50: Misisng `validation.Skip`

regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+		// and must have the correct format.
+		validation.Field(&party.TaxID,
+			validation.Required,
+			validation.By(func(value interface{}) error {
+				tID, ok := value.(*tax.Identity)
+				if !ok || tID == nil {
+					return nil
+				}
+				return validateTaxIdentity(tID)
+			}),
```
samlown, line 74: missing `validation.Skip`

regimes/se/scenarios.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+		// Simplified Tax Invoice
+		{
+			Tags: []cbc.Key{tax.TagSimplified},
+			Note: &tax.ScenarioNote{
+				Key:  org.NoteKeyLegal,
+				Src:  tax.TagSimplified,
+				Text: "Simplified Tax Invoice",
+			},
+		},
```
samlown, line 36: This probably isn't required, as simplified is often implied when customer not present..

regimes/se/scenarios.go
```diff
@@ -0,0 +1,38 @@
...
+		// It can never happen for national operations (i.e. both parties are registered in Sweden,
+		// whether B2B or B2C).
+		{
+			Tags: []cbc.Key{tax.TagReverseCharge},
+			Note: &tax.ScenarioNote{
+				Key:  org.NoteKeyLegal,
+				Src:  tax.TagReverseCharge,
+				Text: "Reverse charge: Customer to account for VAT to the relevant tax authority.",
+			},
+		},
```
samlown, line 27: This more generic implementation is also available in the `common.InvoiceScenarios()` method.

regimes/se/tax_identities.go
```diff
@@ -0,0 +1,79 @@
...
+
+// normalizeTaxIdentity performs normalization specific to Swedish tax IDs,
+// ensuring the code is normalized and the country prefix is added if missing.
+func normalizeTaxIdentity(id *tax.Identity) {
+	tax.NormalizeIdentity(id)
+	// TO-DO: decide if this is necessary. If not, we may need to remove the check digit suffix.
+	// Re-add the SE prefix if missing.
+	if id.Code.String()[:2] != taxCodeCountryPrefix {
+		id.Code = cbc.Code(taxCodeCountryPrefix + id.Code.String())
+	}
```
samlown, line 37: In GOBL, all tax identities are normalized **without** the leading country code. The `tax.Identity`'s `String()` method will always add the country code up front.

regimes/se/tax_identities.go
```diff
@@ -0,0 +1,79 @@
...
+// Assumes the code has already been normalized.
+func validateTaxIdentity(tID *tax.Identity) error {
+	return validation.ValidateStruct(tID,
+		validation.Field(&tID.Code,
+			validation.By(validateTaxCode),
+		),
+	)
+}
+
+// PERF: validate this is indeed faster than a RegEx.
```
samlown, line 50: Performance isn't a high concern in GOBL. There is already a large amount of complexity so a pre-compiled Regexp has negligible impact.

regimes/se/tax_categories.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/invopop/gobl/tax"
+)
+
+// Tax rates specific to Sweden.
+const (
+	TaxRateStandard       cbc.Key = "Normalskattesats"
+	TaxRateReduced        cbc.Key = "Skattesats 12 procent"
+	TaxRateHeavilyReduced cbc.Key = "Skattesats 6 procent"
+	TaxRateExempt         cbc.Key = "Momsfri"
+)
```
samlown, line 17: Not quite sure what these are for... They'd also fail validation as `cbc.Key` has a strict format.

regimes/se/org_identities.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+// These are placeholder keys and may be changed according to the needs of the project.
+const (
+	// IdentityKeyOrgNr defines the key for the Swedish Organization Number (Organisationsnummer).
+	IdentityKeyOrgNr cbc.Key = "se-orgnr"
+	// IdentityKeyPersonNr defines the key for the Swedish Person Number (Personnummer).
+	IdentityKeyPersonNr cbc.Key = "se-personnr"
+	// IdentityKeyCoordinationNr defines the key for the Swedish Coordination Number (Samordningsnummer).
+	IdentityKeyCoordinationNr cbc.Key = "se-coordnr"
```
samlown, line 22: The pattern we're going for recently is to use codes as opposed to keys. I'd assume there are local codes for each identity like `ON` or `PN` or something that is used frequently.

regimes/se/tax_categories.go
```diff
@@ -0,0 +1,94 @@
...
+	{
+		Code: tax.CategoryVAT,
+		Name: i18n.String{
+			i18n.EN: "VAT",
+			i18n.SE: "Moms",
+		},
+		Title: i18n.String{
+			i18n.EN: "Value Added Tax",
+			i18n.SE: "Moms",
+		},
```
samlown, line 29: Often a good idea to list `Sources` here so that we know where to look up data in the future.

regimes/se/tax_categories.go
```diff
@@ -0,0 +1,94 @@
...
+						Percent: num.MakePercentage(250, 3),
+						Since:   cal.NewDate(1990, 1, 1),
+					},
+				},
+			},
+			{
+				Key: tax.RateReduced,
+				Name: i18n.String{
+					i18n.EN: "First Reduced Rate",
+					i18n.SE: "Skattesats 12 procent",
```
samlown, line 49: Probably don't want numbers in the names as they may change in the future.

regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/gobl/tax"
+	"github.com/invopop/validation"
+)
+
+// invoiceValidator adds validation checks to invoices which are relevant
+// for the region.
+type invoiceValidator struct {
+	inv *bill.Invoice
+}
```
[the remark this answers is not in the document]
BoscoDomingo: What's the suggested approach? I'll make sure to include that in the contributing guide

regimes/se/invoices.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil
+	}
+
+	return validation.ValidateStruct(party,
+		validation.Field(&party.Name,
+			validation.Required,
+		),
+		validation.Field(&party.Addresses,
+			validation.Required,
+		),
```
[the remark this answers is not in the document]
BoscoDomingo: If I understood the specs correctly, they do not give any requirements for the address itself, only that it be present. In fact, neither party need be located in Sweden, only (at least one of them) registered for tax. I may be mistaken, however!

regimes/se/tax_categories.go
BoscoDomingo: Vestigial from copy-pasting from other countries' regimes. Removing

regimes/se/scenarios.go
BoscoDomingo: Yeah in that case I'll simply use the generic `InvoiceScenarios` as you pointed out. I was in between using it directly and expanding it with the `TagSimplified` and this. Followed what other regimes did for consistency.

regimes/se/tax_categories.go
BoscoDomingo: Fair point. I'll stick to the translation from the English term then (current ones come directly from the Swedish Tax Agency's website)

regimes/se/tax_identities.go
BoscoDomingo: Then the question is if that would be easier to understand than a few `if` statements. Since Swedish tax code validation is so simple, I'd say this approach makes it more maintainable, but open to feedback!

regimes/se/tax_identities.go
BoscoDomingo: My only concern is that in the case of other countries (for example Spain) that prefix is optional (you could just include a business' CIF without `ES` and still be valid). However in Sweden, it **must** always be there, hence why I'm on the fence. Should I still remove it regardless?

regimes/se/org_identities.go
BoscoDomingo: Let me get back to you on that

regimes/se/invoices.go
BoscoDomingo: Right, that's in the `decisions.md` file since I wasn't sure whether the net balance of adding a `Skip` could be too detrimental to correctness for the cost in performance. Thanks for clarifying!

regimes/se/invoices.go
BoscoDomingo: This is another one of the doubts I added to `decisions.md` as I'm not fully clear on differences between `tax.Identity` vs `org.Identity` and which cases require which. You are right that we only need the VAT ID. However, the actual validation for whether that code is valid is really done in the `org.Identity` where I check the format, numbers and Luhn algorithm all check out.

It sounds to me like I should move those here and remove the `org` identities?

internal/luhn.go
BoscoDomingo: I'll move the Luhn tests there since I do see value in having those

BoscoDomingo: And as an extra: the Luhn algorithm being something so general (it can be used outside of regimes, such as for validating credit card numbers) should it still reside in `regimes/common` or is `internal` a better place?

BoscoDomingo:
@samlown How can I get the `out` directory in examples? I thought it was autogenerated

regimes/se/invoices.go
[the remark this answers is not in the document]
samlown: The question is which specs you're looking at. Tax regimes should reflect what is defined in the law as opposed to any specific format, its very unlikely that the law requires customer address details, but if it does, a comment that links to that requirement would be very useful.

regimes/se/invoices.go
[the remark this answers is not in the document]
samlown: Functional methods. France is a good example: https://github.com/invopop/gobl/blob/main/regimes/fr/invoices.go

If there is a context, then that should be probably be maintained. A supplier inside an invoice is a specific context.

regimes/se/invoices.go
samlown: Org Identities can be really useful to have with validation in case a user whats to add them to an invoice, but they are secondary to the Tax Identity which is the main code used for international transactions. Most countries, with the notable exception of the US have a Tax ID.

regimes/se/tax_identities.go
samlown: If you remove the comment, no one will question it ;-)

regimes/se/org_identities.go
samlown: No need to prefix with the country code. Most cases are already context aware. France has a good example of this: https://github.com/invopop/gobl/blob/main/regimes/fr/fr.go#L16-L22

samlown:
> @samlown How can I get the `out` directory in examples? I thought it was autogenerated

Out directories are generated with the `go generate .` command.

regimes/se/invoices.go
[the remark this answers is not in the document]
BoscoDomingo: For regular invoices, yes, [both customer and supplier addresses are necessary](https://www4.skatteverket.se/rattsligvagledning/edition/2025.2/394342.html#h-Saljarens-och-koparens-fullstandiga-namn-och-adress). They [explicitly ruled it](https://www4.skatteverket.se/rattsligvagledning/415583.html?date=2022-09-20), it seems (both in Swedish).

For simplified invoices, [customer is omitted and supplier can be identified exclusively by VAT ID](https://www4.skatteverket.se/rattsligvagledning/28101.html?date=2005-06-27) (in Swedish)

regimes/se/tax_identities.go
BoscoDomingo: The voice in the back of my mind certainly will... But I can live with that 😆

BoscoDomingo:
> > @samlown How can I get the `out` directory in examples? I thought it was autogenerated
> 
> Out directories are generated with the `go generate .` command.

I may be missing something. I also tried in `main` and it didn't work for the `es` regime either (I deleted its `out` and it wasn't regenerated).

samlown:
`out` folders are not created: https://github.com/invopop/gobl/blob/main/examples_test.go#L114

BoscoDomingo:
> `out` folders are not created: https://github.com/invopop/gobl/blob/main/examples_test.go#L114

Even when the `out` directory is there, files aren't generated, even in `main` with `go generate .`.

Running `go test ./examples_test.go --update` however, did create them. Is that the correct way? And obviously this must be added to the contribution guide and probably the README too

BoscoDomingo:
@samlown What's left for this to be merged?

regimes/common/common.go
```diff
@@ -1,5 +1,5 @@
 // Package common provides re-usable regime related structures and data.
-package common
+package common //nolint:revive
```
samlown: I'd avoid `nolint` at the package level. If could be hiding important details. This comment applies to the other files that were modified in the package.

internal/luhn.go
samlown: I see your point, I think the main issue however is polluting the `internal` package with utility functions. Until we have more use-cases of this, I'd suggest sticking with the current `common` package associated with Regimes.

An alternative approach would be to create a package in `pkg` called `checksum` or similar that provides these utility methods.

Another approach would be to incorporate a validation Rule inside the `cbc` package alongside the `Code` type to specifically check the luhn check digit.

regimes/se/decisions.md
samlown, line 1: This one is a bit long, but its probably best inside the PR itself rather than something permanent in the repo.

regimes/se/org_identities.go
```diff
@@ -0,0 +1,194 @@
...
+var (
+	// ValidOrgIdentityTypes defines the keys for the Swedish organization identities.
+	ValidOrgIdentityTypes = []cbc.Code{IdentityTypeOrgNr, IdentityTypePersonNr, IdentityTypeCoordinationNr}
+
+	// ErrInvalidOrgNumberFormat indicates that the organization number is not in the correct format
+	ErrInvalidOrgNumberFormat = errors.New("invalid organization number format")
+	// ErrInvalidIndividualNumberFormat indicates that the person or coordination number is not in the correct format
+	ErrInvalidIndividualNumberFormat = errors.New("invalid person or coordination number format")
+	// ErrInvalidChecksum indicates that the identification number's checksum is invalid
+	ErrInvalidChecksum = errors.New("invalid identification number checksum")
```
samlown: I'd question if its a good idea or not to include these as constants... these are specific validation errors what will require human intervention to solve. The value of an associated Go error type is extremely limited, as the software cannot know what to do if any of them occur. I'd suggest removing them, and ensuring the `validation` errors are correctly assigned.

regimes/se/org_identities.go
```diff
@@ -0,0 +1,194 @@
...
+				if c >= '0' && c <= '9' {
+					digitsOnly += string(c)
+				}
+			}
+
+			// If we don't have the expected number of digits, it's likely not valid and no safe operation
+			// can be performed.
+			if len(digitsOnly) != taxCodeLengthWithoutCheckDigits {
+				return
+			}
```
samlown, line 119: This one looks like it should be tested, as reported by golanglint.

regimes/se/org_identities.go
```diff
@@ -0,0 +1,194 @@
...
+//
+// If the number is not valid, it returns an error.
+//
+// If the organization type is not valid, it returns nil.
+func validateOrgIdentity(id *org.Identity) error {
+	// No need for nil check, as the identity type is validated before this function is called
+
+	return validation.ValidateStruct(id,
+		validation.Field(&id.Code,
+			validation.By(func(value any) error {
```
samlown, line 148: I'd split this validation into its own function for clarity and an appropriate name.

regimes/se/org_identities.go
```diff
@@ -0,0 +1,194 @@
...
+				// for _, c := range code.String() {
+				//  if c >= '0' && c <= '9' {
+				//    digitsOnly += string(c)
+				//  }
+				// }
+
+				switch id.Type {
+				case IdentityTypeOrgNr:
+					if !orgNrRegex.MatchString(digitsOnly) {
+						return ErrInvalidOrgNumberFormat
```
samlown: As mentioned previously, I'd just stick with `errors.New("...text...")` rather than assigning a type.

regimes/se/tax_categories.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/invopop/gobl/cal"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/tax"
+)
+
+var taxCategories = []*tax.CategoryDef{
+	{
+		// Source (English): https://www.skatteverket.se/servicelankar/otherlanguages/inenglishengelska/businessesandemployers/startingandrunningaswedishbusiness/declaringtaxesbusinesses/vat/vatratesandvatexemption.4.676f4884175c97df419255d.html
+		// Source (Swedish): https://www4.skatteverket.se/rattsligvagledning/394966.html
```
samlown: `tax.CategoryDef` has a `Sources` property. I think all these links could go there.

regimes/common/common.go
samlown: My bad here. I see this is a recent change to the linter. I'd be inclined to question if `common` is the most appropriate name now... we can keep the `nolint:revive` comment for the time being.

BoscoDomingo: I kind of agree with the linter, however, I don't quite see any other useful name for such a package (plus it'd be quite tangential to this change) so I went for the silencing instead.

Happy to bounce ideas for a name elsewhere 👍🏻

regimes/se/decisions.md
BoscoDomingo: Right, this one was just for interview purposes. More than happy to remove since it no longer serves much of a purpose (I can maybe add some bits to the README)

regimes/se/org_identities.go
BoscoDomingo: IIRC I took this pattern from elsewhere in the code and while useful for testing (checking not just any error is thrown) they can _probably_ be removed.

> and ensuring the validation errors are correctly assigned.

~Mind if I ask what exactly you mean by this?~ Nvm, saw your comment further down

regimes/se/org_identities.go
BoscoDomingo: The main problem is that 2 fields in this struct (`Code` and `Type`) are intertwined. Without an inlined anonymous function, I'm not sure I can access both at once (at least not from what I've seen in the `validation` package)

BoscoDomingo: It results in something like this:

```go
return validation.ValidateStruct(id,
	validation.Field(&id.Code,
		validation.By(func(value any) error {
			return validateOrgIdentityCode(value, id)
		}),
		validation.Skip,
	),
)
```

regimes/se/org_identities.go
BoscoDomingo: Removed, although mind that the tests are easier to break now since a change in the error message would need to be propagated to all tests that expect it; it will no longer be automatic

BoscoDomingo:
> Finally ready to merge this one! Many thanks @BoscoDomingo!

Love to see it!! Glad I could be of help <3

PR #519, opened 2025-04-27
Boy Scout from adding Swedish regime

BoscoDomingo:
ℹ️ This PR comes from changes made during work on #518

- Fix a small typo in `es`
- Added a first draft of a `CONTRIBUTING.md` guide (feedback _particularly_ welcome)
- Improved `MakePercentage` documentation
- Made the dependency on golanci-lint explicit and more easily configurable via `tools.go`
- Added a template regime (WIP) to ease adding new regimes. Pending more documentation, examples, explanations, etc

## Pre-Review Checklist

- [X] I've modified or created example GOBL documents to show my changes in use, if appropriate.
- [X] When adding or modifying a tax regime or addon, I've added links to the source of the changes, either structured or in the comments.
- [X] I've run `go generate .` to ensure that the Schemas and Regime data are up to date.


samlown (changes requested):
Great suggestions here! Thanks! We've been meaning to get a CONTRIBUTING.md guide in there for some time, so special thanks for that :-)

It'd be great if this can be separated out from the Swedish PR.

BoscoDomingo:
> Great suggestions here! Thanks! We've been meaning to get a CONTRIBUTING.md guide in there for some time, so special thanks for that :-)
> 
> It'd be great if this can be separated out from the Swedish PR.

Sure thing, I'll move the CountryMap and decisions.md over to #518 to keep this one Sweden-free 👍🏻

mise.toml
```diff
@@ -0,0 +1,3 @@
+[tools]
+go = "1.23.0"
+golangci-lint = "1.64.8"
```
samlown, line 3: This feels a bit too opinionated for an open source project. I'd rather avoid having to keep versions in sync in multiple locations. If it can't be updated automatically, probably best in the .gitignore file.

go.mod
```diff
@@ -5,47 +5,216 @@ go 1.23.0
...
 require (
-	cloud.google.com/go v0.110.2
+	cloud.google.com/go v0.116.0
 	github.com/LastPossum/kamino v0.0.2
-	github.com/Masterminds/semver/v3 v3.2.1
+	github.com/Masterminds/semver/v3 v3.3.0
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2
 	github.com/go-jose/go-jose/v4 v4.0.5
-	github.com/google/go-cmp v0.6.0
+	github.com/golangci/golangci-lint v1.64.8
```
samlown, line 13: This seems to add a massive amount of sub-dependencies, I'm not convinced... is there a link to materials where this is suggested as best practice?

samlown (commented):
Much better just on its own, thanks :-) I'm not sure what the advantages of the linter being embedded inside go.mod are though.

go.mod
BoscoDomingo: TIL since Go 1.24 the [`tools` directive](https://go.dev/doc/go1.24#tools) is available for `go.mod`. If you decide to upgrade, then that would be the official way. As you can see, there they mention the use of `tools.go` as the previous workaround. For reference, [here's the comment I mentioned in our chat](https://github.com/golang/go/issues/25922#issuecomment-413898264) where it was regarded as the best approach back in 2018.

In any case, since Go 1.17 it seems that [any modules not needed by the `main` package are automatically pruned out](https://go.dev/ref/mod#graph-pruning) (if I understood correctly) so either way it _shouldn't_ affect build or test times.

LMK what your preference is, I'll adapt the PR accordingly!

samlown: Okay, thanks for that, makes more sense now. I think what's lacking however is an example of how this is actually used. We have examples for Go generate, but its not clear how to use the tools.

We can upgrade to 1.24 for this public package as its not nice to force upgrades. Its already on 1.23 due to an external dependency.

CHANGELOG.md
```diff
@@ -6,6 +6,20 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
...
 ## [Unreleased]
 
+### Added
+
+- `regime`: added a template regime.
+
+### Changed
+
+- `tools`: add explicit dependency for golangci-lint
+- `docs`: add `CONTRIBUTING.md` file
```
samlown, line 16: These are additions as opposed to changes. We use the backticks for package names, I don't think that's needed here.

go.mod
BoscoDomingo: > We can upgrade to 1.24 for this public package as its not nice to force upgrades. Its already on 1.23 due to an external dependency.

It's not clear to me if you are saying we _can_ or _can't_ upgrade. I assume it's the latter, but since Go versions are backwards-compatible, [upgrading minor versions should never break](https://tip.golang.org/doc/go1compat).

And as for the example I'll do it tomorrow after work, should be fairly easy!

PR #526, opened 2025-05-07
Additional Portugal SAF-T Validations

cavalle:
Adds validation and normalization logic to `pt-saft-v1` to prevent non-compliant invoices to be valid. Namely:

- Validates advances' date is present and matches the invoice issue date. 
  - Normalises advances to set the invoice issue date when nil
- Validates due total in invoice-receipts is zero (they must be fully paid)
- Validates preceding issue date is not after the invoice issue date
- Validates the total with taxes in simplified invoices doesn't exceed the legal maximum of 1000€

## Pre-Review Checklist

### DIFF
--- a/addons/pt/saft/invoices.go
+++ b/addons/pt/saft/invoices.go
@@ -7,8 +7,10 @@ import (
 	"strings"
 
 	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/cal"
 	"github.com/invopop/gobl/cbc"
 	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/org"
 	"github.com/invopop/gobl/pay"
 	"github.com/invopop/gobl/tax"
 	"github.com/invopop/validation"
@@ -63,6 +65,13 @@ func validateInvoice(inv *bill.Invoice) error {
 		),
 		validation.Field(&inv.Totals,
 			validation.By(validateTotals(inv)),
+			validation.Skip,
+		),
+		validation.Field(&inv.Preceding,
+			validation.Each(
+				validation.By(validatePreceding(inv)),
+				validation.Skip,
+			),
 			validation.Skip,
 		),
 	)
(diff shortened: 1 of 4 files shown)


addons/pt/saft/invoices.go
```diff
@@ -26,12 +30,18 @@ var (
...
 var invoiceWorkTypes = []cbc.Code{
 	WorkTypeProforma,
 	WorkTypeConsignmentInv,
 	WorkTypeConsignmentCredit,
 }
 
+// simplifiedMaxAmount is the maximum amount allowed for simplified invoices according to
+// Article 40 of the Portuguese VAT Code. This is the most permissive limit (for retail
+// sales) since we cannot distinguish between retail and non-retail sales.
+var simplifiedMaxAmount = num.MakeAmount(1000, 0)
```
samlown, line 43: I've always tried to avoid these types of validations in GOBL that may change over time or have specific use cases, without a specific structure to deal with them like tax rates.

Appreciating that this is something that is being requested, I wonder if it could be handled by "scenarios" with some extra tweeks, at least then tags could be used to distinguish between retail and non-retail scenarios perhaps.

addons/pt/saft/invoices.go
```diff
@@ -55,9 +65,80 @@ func validateInvoice(inv *bill.Invoice) error {
...
+func validateTotals(inv *bill.Invoice) validation.RuleFunc {
+	return func(val any) error {
+		tot, _ := val.(*bill.Totals)
+		if tot == nil {
+			return nil
+		}
+
+		return validation.ValidateStruct(tot,
+			validation.Field(&tot.Due,
+				validation.When(isInvoiceReceipt(inv), num.Max(num.AmountZero)),
```
samlown, line 131: Should this not be an Equals zero?

samlown (approved):
Happy to merge this. I think fixing the amount maximums in the code will bite us in the future.

addons/pt/saft/invoices.go
cavalle: In fact, in the last round of feedback from the certifiers, they complained that we hadn't apply the 100 eur limit. Let's discuss whether it's worth distinguishing between retail and non-retail and how to better enforce this constraint (via validations or scenarios).

addons/pt/saft/invoices.go
cavalle: yes, but I don't think we have a `num.Equals` or `validation.Equals` validation at the moment. Based on what we have at the moment, it was either `number.Max` or `validation.In`. Would you prefer `validation.In` (if it works with amounts) or should we add a `num.Equals` validation function?

samlown: BTW, there is now a `num.Equals` validator, and a few others.

cavalle:
Outdated

PR #527, opened 2025-05-09
Fix imprecise due total calculation

cavalle:
- Performs all payable totals calculations after rounding.
- Fixes issue with pre-rounding of tax totals.
- Ensures that advances are always rounded to the precision of the currency before calculations.

## Pre-Review Checklist

- _N/A_ I've modified or created example GOBL documents to show my changes in use, if appropriate.
- _N/A_ When adding or modifying a tax regime or addon, I've added links to the source of the changes, either structured or in the comments.
- _N/A_ I've run `go generate .` to ensure that the Schemas and Regime data are up to date.

### DIFF
--- a/bill/calculator.go
+++ b/bill/calculator.go
@@ -165,7 +165,7 @@ func calculate(doc billable) error {
 		t.Total = t.Total.Subtract(ti)
 	}
 
-	// Finally calculate the total with *all* the taxes.
+	// Calculate the total with *all* the taxes.
 	t.Tax = t.Taxes.Sum
 	t.TotalWithTax = t.Total.Add(t.Tax)
 	if t.Taxes.Retained != nil {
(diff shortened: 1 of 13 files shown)


bill/invoice_test.go
```diff
@@ -685,9 +729,10 @@ func TestRemoveIncludedTaxDeep(t *testing.T) {
 		t.Logf("TOTALS: %v", string(data))
 	*/
 
+	assert.Equal(t, "1069.81", i.Totals.Tax.String())
 	assert.Equal(t, "17830.19", i.Totals.Total.String())
 	assert.Equal(t, "17830.20", i2.Totals.Total.String())
-	assert.Equal(t, "-0.01", i2.Totals.Rounding.String())
+	assert.Equal(t, "-0.02", i2.Totals.Rounding.String())
```
samlown, line 735: I'd broken this one in an earlier PR 🤦

bill/calculator.go
```diff
@@ -175,33 +175,36 @@ func calculate(doc billable) error {
...
+		t.Payable = t.Payable.Add(*t.Rounding)
+	}
 	if pd := doc.getPaymentDetails(); pd != nil {
 		pd.calculateAdvances(zero, t.Payable)
-
-		// Deal with advances, if any
+		// Deal with advances, if any. Note that in the current
+		// implementation multiple percentage advances are likely to
+		// suffer rounding errors. It usually better for users to use
+		// fixed payment amounts if possible.
```
cavalle, line 200: This is very true. And makes me wonder where GOBL should actually be making this calculation. An advance is supposed to be a payment already made and so, already calculated. I wonder if, to ensure no ambiguity can happen, GOBL should only accept an amount (not a percentage) (and only with the currency decimals)

cavalle (commented):
LGTM!

bill/calculator.go
samlown: I think the percent calculation is useful, but it does indeed need a better warning. I'll update the advance object's description to see if that helps. 👍

PR #531, opened 2025-05-13
Extend and export the billable interface

cavalle:
- Extend and export the methods in the `billable` interface to allow for generic handling of bill documents (Invoice, Payment, Delivery and Order) outside of the package.

## Pre-Review Checklist

- [ ] N/A I've modified or created example GOBL documents to show my changes in use, if appropriate.
- [ ] N/A When adding or modifying a tax regime or addon, I've added links to the source of the changes, either structured or in the comments.
- [ ] N/A I've run `go generate .` to ensure that the Schemas and Regime data are up to date.

### DIFF
--- a/bill/payment.go
+++ b/bill/payment.go
@@ -319,6 +319,123 @@ func (pmt *Payment) calculate() error {
 	return nil
 }
 
+/** Billable interface implementation **/
+
+// GetSeries returns the series of the payment.
+func (pmt *Payment) GetSeries() cbc.Code {
+	return pmt.Series
+}
+
+// GetCode returns the code of the payment.
+func (pmt *Payment) GetCode() cbc.Code {
+	return pmt.Code
+}
+
+// GetIssueDate returns the issue date of the payment.
+func (pmt *Payment) GetIssueDate() cal.Date {
+	return pmt.IssueDate
+}
+
+// GetIssueTime returns the issue time of the payment.
+func (pmt *Payment) GetIssueTime() *cal.Time {
+	return pmt.IssueTime
+}
+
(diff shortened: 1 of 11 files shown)


bill/payment.go
```diff
@@ -319,6 +319,123 @@ func (pmt *Payment) calculate() error {
...
+
+// GetPaymentDetails returns the payment details of the payment.
+func (pmt *Payment) GetPaymentDetails() *PaymentDetails {
+	return nil // no payment details for deliveries
+}
+
+// GetTotals returns the totals of the payment.
+func (pmt *Payment) GetTotals() *Totals {
+	return nil // no totals for payments
+}
```
samlown, line 402: I don't think this is the best approach here. The `Payment` model does have some of these underlying properties with different types. Providing a getter that just returns nil because the types don't coincide could potentially be very confusing.

samlown (changes requested):
I think this is useful. `GetXX` a bit of an anti-pattern in Go, but there aren't many options when serialising data. I do think the `Payment` is an outlier though, and shouldn't be bundled into the main billable interface. I get the impression there should be more separation between the areas of the document. Perhaps `Header` for all of the base shared fields and a `Detail` interface perhaps for the lines and totals used in Order, Delivery, and Invoice.

cavalle:
Changes rejected in favour of implementing ad-hoc wrappers where necessary, as shown here:  https://github.com/invopop/at-pt/pull/16/files#diff-bbdd19cf28aa931d377c9fe0cff59a8e594e242a4a4ce0eb3f4fd23e5f8b6c58

PR #545, opened 2025-06-11
Refactoring payments and CFDI support for PUE o PPD methods

samlown:
- CFDI addon support for payment method extension, determined automatically using Scenarios.
- Adding Payment fields to handle amount payable, advances, and due.
- Advanced payment tax calculation base on proportions actually paid.

## Pre-Review Checklist

### DIFF
--- a/bill/payment_line.go
+++ b/bill/payment_line.go
@@ -31,6 +31,11 @@ type PaymentLine struct {
 	// installment number for this payment line.
 	Installment int `json:"installment,omitempty" jsonschema:"title=Installment"`
 
+	// Additional human readable description of the payment line which may be useful for
+	// explaining the purpose or special conditions. Notes should be used for more
+	// formal comments.
+	Description string `json:"description,omitempty" jsonschema:"title=Description"`
+
 	// Payable reflects the amount of the document that is payable. This will be
 	// calculated from the embedded document's amount automatically and converted
 	// to the currency of the document.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 - `bill`: `PaymentLine` includes a `refund` boolean to indicate flow of funds, and impacts the `Payment`'s total potentially making it negative.
 - `org`: `DocumentRef` now includes a `payable` property.
-- `bill`: `PaymentLine` includes `payable`, `advances`, `due`, and `tax` properties.
+- `bill`: `PaymentLine` includes `description`, `installment`, `payable`, `advances`, `due`, and `tax` properties.
 - `mx-cfdi-v4`: Define payment method in document.
 
 ## [v0.217.1] - 2025-06-02
--- a/data/schemas/bill/payment.json
+++ b/data/schemas/bill/payment.json
@@ -350,6 +350,11 @@
           "title": "Installment",
           "description": "When making multiple payments for a single document, this specifies the\ninstallment number for this payment line."
         },
+        "description": {
+          "type": "string",
+          "title": "Description",
+          "description": "Additional human readable description of the payment line which may be useful for\nexplaining the purpose or special conditions. Notes should be used for more\nformal comments."
+        },
         "payable": {
           "$ref": "https://gobl.org/draft-0/num/amount",
           "title": "Payable",


addons/pt/saft/payments.go
```diff
@@ -98,8 +97,6 @@ func validatePaymentLine(val any) error {
 			validation.Required,
 			validation.Skip,
 		),
-		validation.Field(&pl.Debit, num.Min(num.AmountZero)),
-		validation.Field(&pl.Credit, num.Min(num.AmountZero)),
```
cavalle, line 102: Shouldn't we validate the amount now? In PT regulations are pretty explicit about preventing documents with negative values from being issued. I added these validations to make sure certifiers can see that the system is doing their best to comply with the law. Should we validate this somewhere else?

bill/payment.go
```diff
@@ -123,12 +122,20 @@ type Payment struct {
 	// to other documents and alternative parties involved in the order-to-delivery process.
 	Ordering *Ordering `json:"ordering,omitempty" jsonschema:"title=Ordering"`
 
-	// Summary of the taxes applied to the payment for tax regimes that require
-	// this information to be communicated.
-	Tax *tax.Total `json:"tax,omitempty" jsonschema:"title=Tax"`
+	// Total advances from any advances registered in the lines. Calculated automatically.
+	Advances *num.Amount `json:"advances,omitempty" jsonschema:"title=Advances,calculated=true"`
```
cavalle: Do we really need to mix together advances from potentially completely unrelated invoices? How this calculation is useful?

bill/payment.go
```diff
@@ -123,12 +122,20 @@ type Payment struct {
...
 	// Total amount to be paid in this payment, either positive or negative according to the
-	// line types and totals.
-	Total num.Amount `json:"total" jsonschema:"title=Total"`
+	// line types and totals. Calculated automatically.
+	Total num.Amount `json:"total" jsonschema:"title=Total,calculated=true"`
+	// Due reflects the amount that is still to be paid and will be calculated automatically
+	// based on the total and advance amounts, and may be require in some tax regimes or addons.
+	Due *num.Amount `json:"due,omitempty" jsonschema:"title=Due,calculated=true"`
```
cavalle: What's the usefulness of summing the due amount of different invoices?

bill/payment_line.go
```diff
@@ -2,92 +2,45 @@ package bill
...
-		validation.Field(&pl.Currency),
-		validation.Field(&pl.Debit,
-			validation.When(
-				pl.Credit == nil,
-				validation.Required.Error("must have either debit or credit"),
-			),
-		),
-		validation.Field(&pl.Credit),
-		validation.Field(&pl.Total, validation.Required),
+		validation.Field(&pl.Installment, validation.Min(1), validation.Max(999)),
```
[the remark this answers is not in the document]
cavalle: This is a good point. If it is optional, we may need to use a pointer here?

org/document_ref.go
```diff
@@ -36,13 +37,18 @@ type DocumentRef struct {
...
 	// Link to the source document.
 	URL string `json:"url,omitempty" jsonschema:"title=URL,format=uri"`
 	// Tax total breakdown from the original document in the provided currency. Should
 	// only be included if required by a specific tax regime or addon.
 	Tax *tax.Total `json:"tax,omitempty" jsonschema:"title=Tax"`
+	// Payable is the total amount that is payable in the referenced document. Only needed
+	// for specific tax regimes or addons. This may be also be used in some scenarios
+	// to determine the proportion of the referenced document that has been paid, and
+	// calculate the remaining amount due and taxes.
+	Payable *num.Amount `json:"payable,omitempty" jsonschema:"title=Payable"`
```
cavalle, line 51: Without context, whether a field from the original document is present in the DocumentRef (e.g. tax or payable) or not (e.g. any other total) feels arbitrary. We added those two fields for ad-hoc reasons and only for payments. This makes me wonder whether those fields really belong here or should belong to the payment itself.

addons/pt/saft/payments.go
samlown: Yes, I've fixes this now and added a `Refund` flag.

bill/payment.go
samlown: Agreed, deleted.

org/document_ref.go
samlown: Its mainly needed for currency conversion. I don't think its that arbitrary though, having the option to provide the amount of the invoice that is payable can be potentially useful in different scenarios. Either way, its optional, and I can't think of a better way to handle the currency conversion side elegantly.

cavalle (approved):

PR #556, opened 2025-07-02
Include validation of latin characters in SDI for names and addresses

pmenendz:
- Added validation of Latin characters in names and addresses. Required by SDI.
- (Linter error) Because common is not an appropriate name 🫠 

## Pre-Review Checklist


addons/it/sdi/invoices_test.go
```diff
@@ -418,3 +418,161 @@ func TestInvoiceLineValidation(t *testing.T) {
 		require.EqualError(t, err, "lines: (0: (item: (name: contains characters outside of Latin and Latin-1 range.).).).")
 	})
 }
```
alvarolivie: I would either move the invalid item name to the func below or move all the Latin1 tests to each individual object like the test above.

pmenendz: Agree, thanks. Moved them to each component. Looks more consisten to me to keep each tests within each object

samlown (approved):
Looks good! Feels like the `validateLatin1String` method should be somewhere it can be shared, but we can leave that for a future validation refactor.

PR #557, opened 2025-07-02
Check price greater than 0

pmenendz:
- mx-cfdi-v4: Validate price is greater than 0 as stated in the specifications:

`ValorUnitario: Si el tipo de comprobante es de “I” (Ingreso), “E” (Egreso) o “N”
(Nómina) este valor debe ser mayor a cero` 

We only support I and E

## Pre-Review Checklist


cavalle (approved):
LGTM!

addons/mx/cfdi/invoice_test.go
```diff
@@ -396,6 +396,28 @@ func TestInvoiceLineItemValidation(t *testing.T) {
...
+			name: "negative price",
+			item: &org.Item{
+				Name:  "Test purchase",
+				Price: num.NewAmount(-5000, 2),
+				Ext: tax.Extensions{
+					cfdi.ExtKeyProdServ: "12345678",
+				},
+			},
+			err: "price: must be greater than 0",
+		},
```
samlown, line 420: Another test is required here for when the price is nil.

samlown (changes requested):
Nil test would be good here. It feels like this should be something more general though.

samlown (approved):
👍

PR #567, opened 2025-07-10
Verifactu: replacement tag

samlown:
- Verifactu: removed customer validation.
- Verifactu: clarifying determination of invoice type
- Verifactu: setting default document type if none provided.

## Pre-Review Checklist

### DIFF
--- a/addons/es/verifactu/bill.go
+++ b/addons/es/verifactu/bill.go
@@ -117,12 +117,19 @@ func validateInvoice(inv *bill.Invoice) error {
 			),
 			validation.When(
 				// Replacement invoices must have a reference to preceding doc.
-				inv.Tax != nil && inv.Tax.Ext.Get(ExtKeyDocType).In("F3"),
+				inv.Tax.GetExt(ExtKeyDocType).In("F3"),
 				validation.Required.Error("details of invoice being replaced must be included"),
 			),
 			validation.Each(
 				validation.By(validateInvoicePreceding(inv)),
 				validation.Skip,
+			),
+			validation.Skip,
+		),
+		validation.Field(&inv.Customer,
+			validation.When(
+				!inv.Tax.GetExt(ExtKeyDocType).In("F2", "R5"), // not simplified
+				validation.Required,
 			),
 			validation.Skip,
 		),
--- a/addons/es/verifactu/bill.go
+++ b/addons/es/verifactu/bill.go
@@ -45,13 +45,20 @@ func normalizeInvoice(inv *bill.Invoice) {
 			ExtKeyCorrectionType: "I",
 		})
 	case bill.InvoiceTypeCorrective:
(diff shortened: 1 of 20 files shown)


addons/es/verifactu/bill.go
```diff
@@ -45,13 +45,20 @@ func normalizeInvoice(inv *bill.Invoice) {
...
-			delete(inv.Tax.Ext, ExtKeyCorrectionType)
+		inv.Tax = inv.Tax.MergeExtensions(tax.Extensions{
+			ExtKeyCorrectionType: "S",
+		})
+	}
+
+	// Set default correction type, unless already provided.
+	switch inv.Type {
+	case bill.InvoiceTypeCreditNote, bill.InvoiceTypeDebitNote, bill.InvoiceTypeCorrective:
+		tx := inv.Tax
```
pmenendz: Why do we need this initialization? inv.Tax is a pointer, right? Then doing this wouldn't be pointing to the same object?

Couldn't we just do inv.Tax.Ext[ExtKeyDocType] = "R1"?

pmenendz (commented):
Just a small comment/ question

addons/es/verifactu/extensions.go
```diff
@@ -383,9 +422,37 @@ var extensions = []*cbc.Definition{
 			i18n.EN: here.Doc(`
 				Identity code used to identify the type of identity document used by the customer.
 
-				Codes "01" and "02" are not defined as they are explicitly inferred from the tax Identity
-				and the associated country. In GOBL, the tax Identity implies association with VAT from
-				Spanish invoices.
+				Codes ~01~, ~02~, and ~07~ are not defined as they are explicitly inferred from the
+				tax Identity and the associated country. ~01~ implies a Spanish NIF, ~02~ is applied
+				when the Tax ID is foreign VAT (or other tax) number, and ~07~ is used when the customer
```
pmenendz: There are several errors in Verifactu that say that the country should be ES when is is 07:

1126 = El valor del CodigoPais solo puede ser ES cuando el IDType sea Pasaporte (03) o No Censado (07). Si IDType es No Censado (07) el CodigoPais debe ser ES.
1233 = Si se identifica a través de la agrupación IDOtro y CodigoPais tiene valor ES, el campo IDType debe valer No Censado (07).

I am testing it

addons/es/verifactu/bill.go
samlown: There was a previous assignment with a new instance, but that's no longer needed. Will fix.

addons/es/verifactu/extensions.go
samlown: You're right Pablo. My conclusion here was that VeriFactu simply does not support invoices without an ID of some kind. I'm not sure how companies in Spain are supposed to invoice US clients, I guess the expectation is just to invent a code or put in `NA` or something.

PR #579, opened 2025-07-22
Fix panic on missing supplier normalizer

cavalle:
- Avoids panic calculating party documents when the regime's normalizer is missing
- Sets normalizer in BR regime definition

## Pre-Review Checklist

### DIFF
--- a/org/party.go
+++ b/org/party.go
@@ -58,7 +58,7 @@ func (p *Party) Calculate() error {
 }
 
 func (p *Party) normalizers() tax.Normalizers {
-	if r := p.RegimeDef(); r != nil {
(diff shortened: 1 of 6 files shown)


org/party.go
```diff
@@ -58,7 +58,7 @@ func (p *Party) Calculate() error {
 }
 
 func (p *Party) normalizers() tax.Normalizers {
-	if r := p.RegimeDef(); r != nil {
+	if r := p.RegimeDef(); r != nil && r.Normalizer != nil {
```
samlown: This feels like it should be handled by a method inside the `tax.Normlizers` as opposed to instantiating an object... I'll have a quick look.

cavalle: Nice one!

PR #580, opened 2025-07-23
Verifactu tax scenarios

pmenendz:
I have created several tax scenarios according with Ley de IVA anc Verifactu rules:
- If the tag is reverse-charge the operation class will be S2 (Verifactu documentation)
- If rate is just exempt, exempt code will be set to E1 (exemptions in internal operations)
- If rate is exempt with the tag export, exempt code will be set to E2 (exemptions in exports of goods)
- If rate is exempt with the tags export and eea, exempt code will be set to E5 (Exemptions in the delivery of goods destined to another Member State.)

### References
- [Verifactu Validation Rules (PDF)](https://www.agenciatributaria.es/static_files/AEAT_Desarrolladores/EEDD/IVA/VERI-FACTU/Validaciones_Errores_Veri-Factu.pdf)
- [Ley del IVA – Articles 21 & 25 (BOE)](https://www.boe.es/buscar/act.php?id=BOE-A-1992-28740)

## Pre-Review Checklist


addons/es/verifactu/tax.go
```diff
@@ -13,15 +13,13 @@ var taxCategoryOpClassMap = tax.Extensions{
 	tax.RateReduced:      "S1",
 	tax.RateSuperReduced: "S1",
 	tax.RateZero:         "S1",
-	tax.RateExempt:       "N1", // General exemption, no tax applies
-	tax.RateExempt.With(tax.TagReverseCharge): "S2",
-	tax.RateExempt.With(tax.TagExport):        "N2",
 }
 
-func normalizeTaxCombo(tc *tax.Combo) {
+func normalizeTaxCombo(tc *tax.Combo, isEuropeanCustomer bool) {
```
samlown: Using boolean flags like this is generally not recommend mainly because its hard to scale.

addons/es/verifactu/verifactu.go
```diff
@@ -67,8 +67,6 @@ func normalize(doc any) {
 	switch obj := doc.(type) {
 	case *bill.Invoice:
 		normalizeInvoice(obj)
-	case *tax.Combo:
-		normalizeTaxCombo(obj)
```
samlown: You need extreme caution when removing normalizations like this. You're effectively making normalization of the tax combos context specific, which is we want to move away from given that there are multiple types of documents.

addons/es/verifactu/bill.go
```diff
@@ -76,6 +76,19 @@ func normalizeInvoice(inv *bill.Invoice) {
 	}
 
 	normalizeInvoicePartyIdentity(inv.Customer)
+
+	for _, line := range inv.Lines {
```
samlown: You've forgotten about document level discounts and charges here.

addons/es/verifactu/extensions.go
```diff
@@ -275,9 +274,13 @@ var extensions = []*cbc.Definition{
 				alongside the ~es-verifactu-op-class~ extension. Values correspond to the
 				L10 list.
 
-				No attempt will be made by GOBL to automatically map tax rates to exemption
-				reason codes, they will need to be determined and applied on a case-by-case
-				basis.
+				There are 2 scenarios where exempt extensions can be automatically assigned:
```
samlown: We shouldn't determine tax combo values based on context, the definitions should be able to handle the different scenarios directly. This is important as Tax Combos need to be atomic.

samlown (changes requested):
Nice try Pablo! I don't however think this is the best approach. Tax Combos need to be self contained and shouldn't depend on outside information in order to operate. This is important to avoid tight coupling.

My recommendation would be to improve the list of rates to be able to cover both exempt and op-class use-cases.

samlown:
This one should now be solved with #584

samlown:
Closing this one as replaced with #584

PR #582, opened 2025-07-29
Portugal Source Billing

cavalle:
- Adds `pt-saft-source-billing` and `pt-saft-source-ref` extensions to handle documents integrated from another system, issued manually or recovered after a data loss.

## Pre-Review Checklist


cavalle:
Working on the test coverage

addons/pt/saft/extensions.go
```diff
@@ -959,6 +968,109 @@ var extensions = []*cbc.Definition{
...
+				your documents. By default, GOBL will set this to "P" (produced in the application).
+
+				The table below shows the supported source billing codes:
+
+				| Code | Name                                    | Description                                    |
+				| ---- | --------------------------------------- | ---------------------------------------------- |
+				| ~P~  | Document produced in the application     | Document generated within the current system   |
+				| ~I~  | Integrated document from another app     | Document integrated from another application   |
+				| ~M~  | Document from recovery or issued manually | Document recovered from backup or issued manually |
+
```
samlown: Sorry, but I'm having real difficulty understanding the naming and the descriptions. "Source Billing" isn't clear to me what is implied and I can't get my head around "Produced" or "Integrated". In the table, the name and the description basically say the same thing, so not much value there.

What I'd suggest for the name is just putting something simple in, like "Produced", and the providing a description for what that actually means and implies.

I appreciate the source XML document uses the `SourceBilling` name though, and that these codes are just weird, but I wonder if there is an easier name. Even just reversing might help so that we end the key with a noun: `billing-source`

samlown (commented):
I find the naming here a bit confusing. I'm wondering if there are any alternatives to make it easier to digest.

addons/pt/saft/extensions.go
cavalle: Agreed the Cursor agent didn't do a great job with the names and descriptions (and I failed to noticed it!). Thanks for pointing it out. I'll fix it. 

As for the name, the documentation calls this field "Origem do document" so I'm gonna try with `pt-saft-doc-source` after that.

cavalle: Addressed in [6f6f0e3](https://github.com/invopop/gobl/pull/582/commits/6f6f0e3e2259d2ffce80f21bc2f43d7729a1e2d4). I finally proposed to use just `source` for the extension, for brevity. LMK!

addons/pt/saft/extensions.go
```diff
@@ -959,6 +968,121 @@ var extensions = []*cbc.Definition{
...
+				Example with a manual document:
+
+				~~~js
+				{
+					"$schema": "https://gobl.org/draft-0/bill/invoice",
+					// ...
+					"ext": {
+						"pt-saft-source-billing": "M",
+						"pt-saft-source-ref": "FTM abc/00001"
```
samlown: I think the example might need updating.

samlown (approved):
Looks good! Just a little comment regarding the examples.

addons/pt/saft/extensions.go
cavalle: Good catch! Fixed (along with a few other cases I found) in d2c5b26

PR #583, opened 2025-08-01
Add fund payments in charges

alvarolivie:
- add support for fund payments as charges in SDI
- add validation for Despatches in SDI

## Pre-Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Added
 
 - `org`: `Item` - `images` field for storing links to images of the item.
+<<<<<<< HEAD
 - `it-sdi-v1`: added fund contributions via charges and validation for despatch (delivery documents)
 - `it`: added new tax category for `CP` to handle code `RT06` for retained taxes
+=======
+- `pt-saft-v1`: positive quantity validation
+- `pt-saft-v1`: tax rate normalization to prevent rate extension and percent mismatches
+- `pt-saft-v1`: exemption note text length validation
+>>>>>>> main
 
 ### Removed
 
--- a/addons/pt/saft/invoices.go
+++ b/addons/pt/saft/invoices.go
@@ -106,7 +106,6 @@ func validateInvoice(inv *bill.Invoice) error {
 		validation.Field(&inv.Lines,
 			validation.Each(
 				bill.RequireLineTaxCategory(tax.CategoryVAT),
-				validation.By(validateInvoiceLine),
 				validation.Skip,
 			),
 			validation.Skip,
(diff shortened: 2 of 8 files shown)


addons/it/sdi/invoices.go
```diff
@@ -51,6 +51,14 @@ func validateInvoice(inv *bill.Invoice) error {
 			),
 			validation.Skip,
 		),
+		validation.Field(&inv.Charges,
+			validation.Each(
+
+				validation.By(validateCharge),
```
samlown: Is the context required? Perhaps you can just validate the `bill.Charge` directly?

addons/it/sdi/invoices_test.go
```diff
@@ -266,6 +266,107 @@ func TestTaxValidation(t *testing.T) {
 	})
 }
 
+func TestChargesValidation(t *testing.T) {
+	ad := tax.AddonForKey(sdi.V1)
+
+	t.Run("Charges with no key", func(t *testing.T) {
```
samlown: Check case on texts for consistency. (start with lower case)

addons/it/sdi/invoices.go
```diff
@@ -239,6 +254,21 @@ func validateAddress(value interface{}) error {
 	)
 }
 
+// validateLatin1String ensures that the item name only contains characters
+// from Latin and Latin-1 range (ASCII 0-127 and extended Latin-1 128-255).
+func validateLatin1String(val any) error {
```
samlown, line 289: I couldn't see any tests for this method.

samlown (changes requested):
Just a few style suggestions. Did you solve the retained tax issue?

addons/it/sdi/invoices.go
alvarolivie: No, you are right, I'll validate directly

addons/it/sdi/invoices.go
alvarolivie: I simply moved this method further down. There should be tests from previous PRs but I'll double check.

addons/it/sdi/invoices_test.go
alvarolivie: 👍

addons/it/sdi/bill.go
```diff
@@ -51,6 +51,10 @@ func validateInvoice(inv *bill.Invoice) error {
 			),
 			validation.Skip,
 		),
+		validation.Field(&inv.Ordering,
+			validation.By(validateInvoiceOrdering(inv.HasTags(TagDeferred))),
```
samlown: I'd move the condition either to a `validation.When` and change the `validateInvoiceOrdering` to be more specific, or modify the method to accept the tags directly and make the decision inside there. As a rule, booleans tend to be hard to trace when making decisions.

alvarolivie: I've simply passed the invoice to the method and checked the tags there. I did not want to check for tags before as the validation is not that simple and requires looking deeper into the object.

PR #584, opened 2025-08-05
Major refactor to include support for new tax Combo key field

samlown:
- Added new `key` field to `tax.Combo`
- Migrate tax `rate` from `standard` to `general`.
- Migrating other exempt tax `rate` values to new `key`.
- Tax key is used to determine specific tax situations, mainly for VAT.
- Advanced extension parsing and mapping utilized in addons to correctly set the tax key.

## Pre-Review Checklist


addons/es/verifactu/extensions.go
```diff
@@ -208,21 +213,16 @@ var extensions = []*cbc.Definition{
...
-				| ~zero~                  | ~S1~            |
-				| ~exempt~                | ~N1~            |
-				| ~exempt+reverse-charge~ | ~S2~            |
-				| ~exempt+export~         | ~N2~            |
+				| Operation Class | Tax Key                |
+				|-----------------|-------------------------|
+				| ~S1~            | ~standard~, ~reduced~, ~super-reduced~, ~zero~ |
+				| ~S2~            | ~reverse-charge~                |
+				| ~N1~            | ~outside-scope~ |
+				| ~N2~            | No key, applied when ~country~ is not ~ES~ |
```
pmenendz, line 225: N2 can also apply when the country is ES, if the transaction is taxed under IGIC (Canary Islands) or IPSI (Ceuta/Melilla) instead of mainland VAT.

For example, clave régimen 08 is used when, due to place-of-supply rules, the transaction isn’t subject to VAT in mainland Spain or the Balearics and may fall under a local tax like IGIC or IPSI. In these cases, N2 is mandatory.

The implementation is fine as it goes in the other direction: If the country is not Spain, then op_class is N2. However, we should take into account that there exist some cases where country can be ES and code N2. For these cases, the user must manually set the extension and the regime code.

Validations document (page 12, 15.6.6): https://www.agenciatributaria.es/static_files/AEAT_Desarrolladores/EEDD/IVA/VERI-FACTU/Validaciones_Errores_Veri-Factu.pdf
Regime 8 description in excel on the _Listas_ page: https://www.agenciatributaria.es/static_files/AEAT_Desarrolladores/EEDD/IVA/VERI-FACTU/DsRegistroVeriFactu.xlsx

addons/es/verifactu/tax_combo.go
```diff
@@ -0,0 +1,124 @@
...
+				Delete(ExtKeyExempt)
+		case tax.KeyOutsideScope:
+			tc.Ext = tc.Ext.
+				Set(ExtKeyOpClass, "N1").
+				Delete(ExtKeyExempt)
+		case tax.KeyExempt:
+			tc.Ext = tc.Ext.
+				SetOneOf(ExtKeyExempt, "E1", "E6").
+				Delete(ExtKeyOpClass)
+		case tax.KeyExport:
```
pmenendz, line 60: Wouldn't we need to map here cases like _exempt+other_ or _export+transport_?

addons/es/verifactu/tax_combo.go
```diff
@@ -0,0 +1,124 @@
...
+
+// prepareTaxComboKey tries to reverse map existing extension keys into the
+// appropriate tax combo key. This helps with the migration period when getting
+// users to move to keys.
+func prepareTaxComboKey(tc *tax.Combo) {
+	if !tc.Key.IsEmpty() {
+		return
+	}
+	switch tc.Ext.Get(ExtKeyExempt) {
+	case "E1", "E6":
```
pmenendz, line 108: Same here. Why don't we map the code E6 to _exempt+other_?

addons/es/verifactu/tax_combo.go
samlown: No, so what I decided to do here (and its quite cool) is allow the user to customize the exempt key if they really need to for these advanced cases, hence the `SetOneOf` method which will allow any of the provided keys.

The key extras get really messy in my opinion as the naming is hard and nuanced.

addons/es/verifactu/extensions.go
samlown: Ah, good point. So in this case the tax combos can be defined manually and would look like:

```json
{
  "taxes": [
    {
      "cat": "VAT",
      "key": "outside-scope",
      "ext": {
        "es-verifactu-exempt": "N2"
      }
    },
    {
      "cat": "IGIC",
      "perecent": "20.0%"
    }
  ]
}
```

This transfers a lot of complexity to the gobl.verifactu convertor which will need to understand that only VAT should be declared and IGIC should be ignored. I don't have a good way to propose how that should be handled at the moment.

PR #591, opened 2025-08-26
Fix regex for frech VAT validation

Haferkorn:
- I updated the REGEX used for validating the 11-chars French VAT. 

In the current state of the code only 11 digits are valid but there can actually be 2 chars. 

> (Eng translation...) The structure of the intra-community VAT number is specific to each country. In France, the number begins with the letters FR, followed by a key (letters or digits assigned by the tax authorities of the company’s registered office location) and ends with the company’s SIREN number (a series of 9 digits).

source: https://www.economie.gouv.fr/cedef/fiches-pratiques/quest-ce-quun-numero-de-tva-intracommunautaire#:~:text=La%20structure%20du%20num%C3%A9ro%20de,(s%C3%A9rie%20de%209%20chiffres).

## Pre-Review Checklist

- [x] I've read the CONTRIBUTING.md guide.
- [x] I have performed a self-review of my code.
- [ ] I have added thorough tests with at **least** 90% code coverage.
- [ ] I've modified or created example GOBL documents to show my changes in use, if appropriate.
- [ ] When adding or modifying a tax regime or addon, I've added links to the source of the changes, either structured or in the comments.
- [ ] I've run `go generate .` to ensure that the Schemas and Regime data are up to date.
- [ ] All linter warnings have been reviewed and fixed.
- [ ] I've been obsessive with pointer nil checks to avoid panics.
- [ ] The CHANGELOG.md has been updated with an overview of my changes.
- [ ] Requested a review from @samlown.

PR #592, opened 2025-09-01
Support for Tax Identity exceptions in France

samlown:
- Replaces #591 
- Handles exception cases in French Tax IDs for Government and La Poste codes.

## Pre-Review Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have added thorough tests with at **least** 90% code coverage.
- [ ] I've run `go generate .` to ensure that the Schemas and Regime data are up to date.
- [ ] The CHANGELOG.md has been updated with an overview of my changes.

### DIFF
--- a/regimes/fr/tax_identity.go
+++ b/regimes/fr/tax_identity.go
@@ -12,8 +12,15 @@ import (
 )
 
 var (
-	taxCodeVATRegexp   = regexp.MustCompile(`^\d{11}$`)
+	taxCodeVATRegexp   = regexp.MustCompile(`^[A-Z\d]{2}\d{9}$`)
+	taxCodeWithLetters = regexp.MustCompile(`[A-Z]`)
 	taxCodeSIRENRegexp = regexp.MustCompile(`^\d{9}$`)
+)
+
+const (
+	// Special exception case for La Poste which doesn't use the same
+	// Luhn check. Source: https://fr.wikipedia.org/wiki/Formule_de_Luhn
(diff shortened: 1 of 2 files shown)


regimes/fr/tax_identity.go
```diff
@@ -52,11 +59,19 @@ func validateVATTaxCode(value interface{}) error {
...
 	}
 
-	// Extract the last nine digits as an integer.
-	siren := str[2:] // extract last nine digits
-	chk := calculateVATCheckDigit(siren)
-	expectStr := str[:2] // compare with first two digits
-	if chk != expectStr {
+	chk := str[:2]   // first 2
+	siren := str[2:] // last 9
+	if taxCodeWithLetters.MatchString(chk) {
```
alvarolivie, line 64: What if the tax code starts with one letter and one number. I'm not sure if it's possible but https://github.com/ltns35/go-vat/blob/main/countries/france.go covers it. I would do the opposite and check if the first two digits are numbers.
```suggestion
	if !taxCodeWithDigits.MatchString(chk) {
```

alvarolivie (commented):
I'd double check what I've commented as I'm not 100% sure. The rest looks good.

regimes/fr/tax_identity.go
samlown: The combination does not matter, its only relevant that the there is a letter somewhere, hence the proposed regexp that just detects a letter anywhere in the string.

PR #600, opened 2025-09-09
Preparing release candicate 0.300-0-rc2

samlown:
Slightly extending the meaning of what a normal Release Candidate would look like with some significant changes for PT.


cavalle (approved):
Nice!

PR #610, opened 2025-09-18
SAFT PT Extra validations and normalizations for compliance

cavalle:
- Validate line quantities are greater than 0
- Normalize tax combo rate to avoid mismatches between the percent and the SAF-T tax rate
- Validate exemption note test minimum length of 5 characters

## Pre-Review Checklist

### DIFF
--- a/addons/pt/saft/line.go
+++ b/addons/pt/saft/line.go
@@ -7,6 +7,7 @@ import (
 	"github.com/invopop/gobl/bill"
 	"github.com/invopop/gobl/cbc"
 	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/num"
 	"github.com/invopop/gobl/org"
 	"github.com/invopop/gobl/tax"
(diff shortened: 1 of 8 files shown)


addons/pt/saft/line.go
```diff
@@ -52,13 +54,33 @@ func validateLine(line *bill.Line) error {
...
+			),
+		),
 		validation.Field(&line.Notes,
 			validation.By(validateLineNotes(line)),
 			validation.Skip,
 		),
 	)
 }
 
+func validateInvoiceLineDiscount(val any) error {
```
samlown: I think this method should reflect the type, so it can potentially be re-used:

```suggestion
func validateBillLineDiscount(val any) error {
```

Does the same not need to be applied to charges?

cavalle: Charges aren't supported in Portugal SAF-T. I renamed the function in ab4b77b07aa6a5c5fe35f7bf7356efd6cd2226dd

PR #630, opened 2025-10-27
Brazil NF-e v4.0 addon

cavalle:

## Pre-Review Checklist


addons/br/dfe/extensions.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+// Brazilian extension keys for fiscal documents
+const (
+	ExtKeyFiscalIncentive = "br-fiscal-incentive"
+	ExtKeyMunicipality    = "br-municipality"
+	ExtKeySimples         = "br-simples"
+	ExtKeySpecialRegime   = "br-special-regime"
+	ExtKeyCFOP            = "br-cfop"
+)
```
samlown, line 12: I'm reluctant to add format specific stuff into regimes. I understand the use-case is for it to be shared between both nfs and nfse. Is there one format that could be used as a base for the second? Its possible to create addon dependencies. I'd assume that there is a main organisation that determines the codes? Perhaps an addon like `br-abrasf-v1` could be the best approach?

The main reason for avoiding complex regimes is because they can be potentially used from other countries, and may change as document versions change. Addons have better support for versions.

addons/br/dfe/identities_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestNormalizeIdentities(t *testing.T) {
+	t.Run("nil identity", func(t *testing.T) {
+		var ident *org.Identity
+		br.Normalize(ident)
+		assert.Nil(t, ident)
+	})
+
+	t.Run("migrates old key br-nfse-municipal-reg", func(t *testing.T) {
+		ident := &org.Identity{
+			Key:  "br-nfse-municipal-reg",
```
samlown: Same here. I'd probably go with a universal: `br-abrasf-municipal-reg`.

addons/br/dfe/extensions.go
cavalle: Moved to `br-dfe-v1` addon

addons/br/dfe/identities_test.go
cavalle: Also moved to `br-dfe-v1`

addons/br/nfe/pay.go
```diff
@@ -0,0 +1,69 @@
...
+	pay.MeansKeyCheque: "02", // Cheque
+	pay.MeansKeyCard:   "03", // Cartão de Crédito
+	pay.MeansKeyDebitTransfer.With(MeansKeyDebit): "04", // Cartão de Débito
+	pay.MeansKeyCreditTransfer:                    "18", // Transferência bancária
+	pay.MeansKeyOnline:                            "18", // Carteira Digital
+	pay.MeansKeyOther:                             "99", // Outros
+}
+
+func normalizePayInstructions(instr *pay.Instructions) {
+	if instr == nil || (instr.Ext != nil && instr.Key == pay.MeansKeyOther) {
```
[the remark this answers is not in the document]
cavalle: Addressed in [3600c9e](https://github.com/invopop/gobl/pull/630/commits/3600c9ef5a2e7df230e8dfe41fbb0363cde7f55a)

addons/br/nfe/pay.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if code := paymentMeansKeyMap[instr.Key]; code != "" {
+		instr.Ext = instr.Ext.Merge(tax.Extensions{
+			ExtKeyPaymentMeans: code,
+		})
+	}
+}
+
+func normalizePayAdvance(adv *pay.Advance) {
+	if adv == nil || (adv.Ext != nil && adv.Key == pay.MeansKeyOther) {
```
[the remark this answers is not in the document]
cavalle: Addressed in [3600c9e](https://github.com/invopop/gobl/pull/630/commits/3600c9ef5a2e7df230e8dfe41fbb0363cde7f55a)

addons/br/dfe/extensions.go
```diff
@@ -0,0 +1,182 @@
...
+				Indicates whether a party benefits from a fiscal incentive.
+			`),
+		},
+		Sources: []*cbc.Source{
+			{
+				Title: i18n.String{
+					i18n.EN: "NFS-e ABRASF Taxpayer Guidance Manual (v2.04)",
+					i18n.PT: "NFS-e ABRASF Manual de Orientação do Contribuinte (v2.04)",
+				},
+				URL: "https://abrasf.org.br/biblioteca/arquivos-publicos/nfs-e-manual-de-orientacao-do-contribuinte-2-04/download",
```
samlown: This suggests that the codes are specific for the NFS-e, is there another source for the NF-e?

addons/br/dfe/extensions.go
```diff
@@ -0,0 +1,182 @@
...
+			},
+		},
```
samlown: Same here, these are all references to NFS-e.

addons/br/dfe/extensions.go
cavalle: @samlown this is a boolean value indicating whether a party qualifies for having a fiscal incentive  according to Lei nº 11.196/2005 (Lei do Bem) (i.e. the company does R&D). A field for this exists both in NFS-e and NF-e and it has the same values ("1" or "2"). I imagine this could also appear in other document types in the future, potentially with different values. 

In GOBL, I'm choosing to follow the ABRASF way of encoding this information and that's why I'm listing it as the only source. We could list both sources, or maybe just the law, instead of the formats, if you think it's better. Let me know.

addons/br/dfe/extensions.go
cavalle: Similar case as before: https://github.com/invopop/gobl/pull/630/files/3600c9ef5a2e7df230e8dfe41fbb0363cde7f55a#r2469221773

addons/br/dfe/extensions.go
samlown: Right, so it probably makes sense to link to both sources if possible so that in the future its easier to fix. I'd be tempted though to have each independent.

cavalle: Ok! On a second (or third, or fourth!) thought, I agree it makes sense to have them independent. I'll make it so.

cavalle: Done

addons/br/dfe/extensions.go
cavalle: Resolved: extensions were splitted

addons/br/nfe/invoices.go
```diff
@@ -0,0 +1,234 @@
...
+			validation.Each(
+				validation.By(validateAddress),
+				validation.Required,
+				validation.Skip,
+			),
+			validation.Skip,
+		),
+		validation.Field(&p.Ext,
+			validation.When(len(p.Addresses) > 0,
+				tax.ExtensionsRequire("br-ibge-municipality"),
```
samlown: This should probably use a constant.

samlown: Fixed it.

PR #635, opened 2025-10-29
IT: supplier phone validation and normalization

pmenendz:
- New SDI validation: In SDI if the supplier is Italian, the telephone number has to be between 5 and 12 characters long.
- New SDI normalization: If the supplier is Italian and the telephone number starts by +39, the prefix is removed.

This last one was giving problems to some clients because the italian number with the prefix is 13 characters long.

## Pre-Review Checklist


addons/it/sdi/bill.go
```diff
@@ -26,6 +26,15 @@ func normalizeSupplier(party *org.Party) {
 		}
 		party.Ext[ExtKeyFiscalRegime] = "RF01" // Ordinary regime is default
 	}
+
+	// Normalize Italian supplier telephone numbers by stripping '+39' prefix
+	if isItalianParty(party) && len(party.Telephones) > 0 {
```
samlown, line 31: Should this not be done for all parties, not just the supplier? You could also in theory normalize directly on all phone numbers, which is probably what you want in this case.

addons/it/sdi/bill.go
```diff
@@ -258,6 +274,20 @@ func validateDespatch(value any) error {
 	)
 }
 
+func validateTelephone(value any) error {
```
samlown: This should either be generic for the org, or have a specific method name pattern to reflect the specific nature. I'd be tempted to just validate *all* phone numbers, and place it in `org.go` or similar.

pmenendz: The SDI validation is only required for supplier telephone numbers and only when they are italian, not customer or other parties.

https://www.fatturapa.gov.it/export/documenti/Technical-Rules-for-European-Invoicing-v2.5.pdf (page 56)

addons/it/sdi/bill.go
pmenendz: mmm, initially I was thinking only about Italian suppliers because sdi only validates the phone for them. However, it might make sense to normalize all italian numbers if the addon is sdi, even the customer's. 

For validation, however, I wouldn't do it general, because SDI only validates the supplier number.

addons/it/sdi/bill.go
```diff
@@ -101,6 +110,13 @@ func validateSupplier(value interface{}) error {
 			validation.Each(validation.By(validateAddress)),
 			validation.Skip,
 		),
+		validation.Field(&supplier.Telephones,
+			validation.When(
+				isItalianParty(supplier) && len(supplier.Telephones) > 0,
```
alvarolivie: I'm not sure if we need the length check, If the length is 0 wouldn't it just skip?

addons/it/sdi/bill.go
```diff
@@ -258,6 +274,20 @@ func validateDespatch(value any) error {
...
 }
 
+func validateTelephone(value any) error {
+	t, ok := value.(*org.Telephone)
+	if !ok {
+		return nil
+	}
+	return validation.ValidateStruct(t,
+		validation.Field(&t.Number,
+			validation.Required,
```
alvarolivie: GOBL already enforces Number as a required field so I think we can remove this.

pmenendz: Right

addons/it/sdi/bill.go
pmenendz: True

PR #644, opened 2025-11-11
Remove detail in favor of notes

alvarolivie:
- Remove detail in favor of notes for payment terms.

## Pre-Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Removed
 
-- `pay`: Deprecated terms.Detail in favour of terms.Notes
+- `pay`: Removed terms.Detail in favour of terms.Notes
 
 ### Fixed
 


pay/terms.go
```diff
@@ -17,7 +17,7 @@ import (
 type Terms struct {
 	// Type of terms to be applied.
 	Key cbc.Key `json:"key,omitempty" jsonschema:"title=Key"`
-	// Text detail of the chosen payment terms.
+	// Text detail of the chosen payment terms (Deprecated).
```
samlown: Rather than adding a "deprecated" label, I'd suggest removing the field directly and adding a `UnmarshalJSON` method that will move an old detail to the notes field.

pay/terms.go
```diff
@@ -92,6 +92,12 @@ func (t *Terms) Normalize() {
 	if t == nil {
 		return
 	}
+
+	if t.Detail != "" && t.Notes == "" {
+		t.Notes = t.Detail
+		t.Detail = ""
+	}
```
samlown: Perhaps if both fields are present the Detail should be added after any existing notes? not sure about this one.

samlown (changes requested):
I'd be better to avoid deprecated fields and just migrate directly.

PR #777, opened 2026-03-31
Add SA regime

migueltorresvalls:
# Saudi Arabia ZATCA Regime

### Regime (`regimes/sa/`)

Country-level tax rules that apply to all Saudi invoices regardless of format:

- **VAT categories** -- Single VAT category with two historical rates: 5% (Jan 2018) and 15% (Jul 2020).
- **Tax identity** -- Validates Saudi VAT IDs: 15 digits starting and ending with `3` (BR-KSA-40).
- **Organization identities** -- 11 identity type codes (TIN, CRN, MOM, MLS, 700, SAG, NAT, GCC, IQA, PAS, OTH) with supplier/customer restrictions (BR-KSA-08).

## Examples

The `examples/sa/` directory contains sample invoices covering:

- Standard invoices (SAR and USD)
- Simplified invoices (including zero-rated with and without notes)
- Credit notes and debit notes (standard and simplified)
- Summary credit notes with delivery periods

## Pre-Review Checklist

- [X] Performed a self-review of my code.
- [X] Added thorough tests with at **least** 90% code coverage.


addons/eu/en16931/tax_combo.go
```diff
@@ -108,9 +109,11 @@ func taxComboRules() *rules.Set {
 		),
 		// BR-S-10, BR-Z-10: standard, zero-rated, IGIC, and IPSI shall NOT have a VATEX code
 		rules.When(is.Func("is non-exempt", taxComboIsNonExempt),
-			rules.Field("ext",
-				rules.Assert("07", "VATEX extension must not be set for standard, zero, IGIC, or IPSI categories (BR-S-10, BR-Z-10)",
-					tax.ExtensionsExclude(cef.ExtKeyVATEX),
+			rules.When(is.FuncContext("regime is not saudi arabia", regimeIsNotSA),
```
alvarolivie: Could this be done with the addon?

addons/sa/zatca/bill.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/cal"
+	"github.com/invopop/gobl/catalogues/cef"
+	"github.com/invopop/gobl/catalogues/untdid"
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/tax"
+)
+
+var taxExemptionReason = map[string]string{
```
alvarolivie: I would set this to cbc.Key: cbc.Code or something along those lines.

addons/sa/zatca/bill_invoices.go
```diff
@@ -0,0 +1,233 @@
...
+
+		rules.Field("issue_time",
+			rules.Assert("01", "issue time must be present (BR-KSA-70)", is.Present),
+		),
+
+		// Tax
+		rules.Field("tax",
+			rules.Assert("02", "tax must be present", is.Present),
+			rules.Field("ext",
+				rules.Assert("03", "extensions keys untdid document type and invoice type transaction are required",
```
alvarolivie: Might be cleaner to split into two rules. Up to you

addons/sa/zatca/bill_invoices.go
```diff
@@ -0,0 +1,233 @@
...
+					tax.ExtensionsRequire(requiredExtensions...)),
+				rules.Assert("04", "document type must be a valid ZATCA type (388, 386, 383, 381) (BR-KSA-05)",
+					tax.ExtensionsHasCodes(untdid.ExtKeyDocumentType, DocTypeTaxInvoice, DocTypePrepayment, DocTypeDebitNote, DocTypeCreditNote),
+				),
+				rules.Assert("05", "invoice transaction type must be valid",
```
alvarolivie: I would always set the code of the rule you are referencing.

addons/sa/zatca/bill_invoices.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func invoiceIsExport(val any) bool {
+	code := getInvTypeCode(val)
+	return code != "" && code[4] == '1'
+}
+
+func invoiceIsSummary(val any) bool {
+	code := getInvTypeCode(val)
+	return code != "" && code[5] == '1'
```
alvarolivie: Any reason to have single and double quotations?

addons/sa/zatca/scenarios.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/catalogues/untdid"
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/tax"
+)
+
+var scenarios = []*tax.ScenarioSet{
+	{
```
alvarolivie: You could also add the rest of the scenarios here right?

addons/sa/zatca/tax_combo.go
```diff
@@ -0,0 +1,103 @@
...
+	VatexIntlTransportRelated     cbc.Code = "VATEX-SA-34-3"
+	VatexQualifyingTransportMeans cbc.Code = "VATEX-SA-34-4"
+	VatexTransportRelated         cbc.Code = "VATEX-SA-34-5"
+	VatexMedicines                cbc.Code = "VATEX-SA-35"
+	VatexQualifyingMetals         cbc.Code = "VATEX-SA-36"
+	VatexPrivateEducation         cbc.Code = "VATEX-SA-EDU"
+	VatexPrivateHealthcare        cbc.Code = "VATEX-SA-HEA"
+	VatexMilitaryGoods            cbc.Code = "VATEX-SA-MLTRY"
+	VatexOutOfScope               cbc.Code = "VATEX-SA-OOS"
+)
```
alvarolivie: You are declaring these to just add them to a map of arrays. I would look into just having 3 arrays and moving the complexity to the case. you are already choosing based on tax category.

addons/sa/zatca/tax_combo_test.go
```diff
@@ -0,0 +1,312 @@
...
+		})
+		require.NoError(t, inv.Calculate())
+		err := rules.Validate(inv)
+		assert.ErrorContains(t, err, "VATEX exemption code must be present and valid for Z/E/O categories, and must not be set for Standard")
+	})
+}
+
+// --- Rule 03 (BR-KSA-18): VAT category code must be one of S, Z, E, O ---
+
+func TestTaxComboRule03_ValidCategoryCodes(t *testing.T) {
```
alvarolivie: In general we prefer individual tests and not arrays we iterate. If you find you have duplicated code you can always move some to a function and call it in each test.

regimes/sa/bill_invoices.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/invopop/gobl/rules/is"
+	"github.com/invopop/gobl/tax"
+)
+
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(l10n.SA.Tax())),
+			rules.Field("supplier",
+				rules.Assert("01", "supplier must have a valid tax ID code (BR-KSA-39)",
```
alvarolivie: In this case this rule should not be present as it's the regime which does not have context of the format. If you want another programmer to understand why this is not in the addon you can add a comment there saying that this rule is covered by regime rules. (Up to debate)

regimes/sa/bill_invoices.go
```diff
@@ -0,0 +1,38 @@
...
+					is.Func("valid VAT code", hasTaxIDCode),
+				),
+				rules.Field("identities",
```
alvarolivie: Do they always need an identity. Most countries ask for one or the other. Just double check as it may be that tax law asks for Tax ID or Identity and you only validate the format. Then in the addon you validate that both are present.

regimes/sa/org_identities.go
```diff
@@ -0,0 +1,67 @@
...
+	IdentityTypeIqa      cbc.Code = "IQA" // Iqama Number (Resident ID)
+	IdentityTypePassport cbc.Code = "PAS" // Passport Number
+	IdentityTypeOTH      cbc.Code = "OTH" // Other ID
+)
+
+var (
+	identitiesRegexp = regexp.MustCompile(`^[a-zA-Z0-9]*$`)
+
+	// CustomerValidIdentities groups customer accepted identities by ZATCA
+	CustomerValidIdentities = []cbc.Code{
```
alvarolivie: Again, not sure if the regime should just cover identities and what is allowed. Then extra validation on who should have what lives in the addon.

alvarolivie (commented):
Great work! Just pointed out a few things I would do different. Let me know if you disagree 👍

addons/sa/zatca/bill.go
```diff
@@ -0,0 +1,103 @@
...
+	// BR-KSA-O-01: "Not subject to VAT" lines must have a 0% rate.
+	for _, line := range inv.Lines {
+		vat := line.Taxes.Get(tax.CategoryVAT)
+		if vat == nil {
+			continue
+		}
+		if vat.Key == tax.KeyOutsideScope {
+			vat.Percent = &num.PercentageZero
+		}
+	}
```
alvarolivie: this is handled by gobl.ubl

addons/sa/zatca/bill_invoices.go
migueltorresvalls: code[5] is a byte, so need '' for comparison

addons/sa/zatca/scenarios.go
migueltorresvalls: https://github.com/invopop/gobl/blob/e716a8cdb5668ae184eaf699d9d7862bef738261/tax/addons.go#L99-L108

normalizeAddons() expands each addon into [required, addon], so the addon's scenarios always overwrites the required's.

regimes/sa/org_identities.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/gobl/rules"
+	"github.com/invopop/gobl/rules/is"
+	"github.com/invopop/gobl/tax"
+)
+
+var (
+	identitiesRegexp = regexp.MustCompile(`^[a-zA-Z0-9]*$`)
+)
```
alvarolivie, line 14: Not sure if identity types should be defined here. They are only used in the addon and validation is the same for all of them.

addons/eu/en16931/tax_combo.go
```diff
@@ -108,9 +108,11 @@ func taxComboRules() *rules.Set {
 		),
 		// BR-S-10, BR-Z-10: standard, zero-rated, IGIC, and IPSI shall NOT have a VATEX code
 		rules.When(is.Func("is non-exempt", taxComboIsNonExempt),
-			rules.Field("ext",
-				rules.Assert("07", "VATEX extension must not be set for standard, zero, IGIC, or IPSI categories (BR-S-10, BR-Z-10)",
-					tax.ExtensionsExclude(cef.ExtKeyVATEX),
+			rules.When(is.FuncContext("addon is not ZATCA", addonIsNotZATCA),
```
samlown: I'd leave a comment here explaining where we're using a negative instead of a positive assertion. Probably makes sense to add a specific test for this in the `tax` package alongside the `tax.HasAddon()` method.

addons/sa/zatca/bill.go
```diff
@@ -0,0 +1,93 @@
...
+		inv.Tax = &bill.Tax{}
+	}
+
+	// Always set rounding to currency for SA ZATCA
+	inv.Tax.Rounding = tax.RoundingRuleCurrency
+
+	// Ensure issue time exists
+	if inv.IssueTime == nil {
+		inv.IssueTime = &cal.Time{}
+	}
```
samlown: This would default to 00:00:00, are you sure this is what you want?

addons/sa/zatca/bill.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// BR-KSA-83
+	// VAT categories E,Z,O must have an associated tax note. This
+	// validation adds them if not previously provided by the user
+	for _, line := range inv.Lines {
```
samlown: For sure this should be in its own method for clarity.

examples/sa/simplified-zero-rated-with-notes.yaml
```diff
@@ -0,0 +1,64 @@
...
+    quantity: "1"
+    item:
+      name: "Medicine | الدواء"
+      price: "50.00"
+      unit: "PCE"
+    taxes:
+      - cat: VAT
+        rate: zero
+        ext:
+          untdid-tax-category: "Z"
```
samlown: Wouldn't this be normalized?

examples/sa/simplified-zero-rated-with-notes.yaml
```diff
@@ -0,0 +1,64 @@
...
+uuid: "8d487816-70b8-4ade-a618-9d620b73814a"
+issue_date: "2022-09-07"
+issue_time: "12:21:28"
+code: "123456"
+regime: "SA"
+
+tax:
+  ext:
+    sa-zatca-invoice-type: "0200000"
+    untdid-document-type: "388"
```
samlown: These can't be determined automatically?

examples/sa/standard-invoice.yaml
samlown, line 1: Just in general for all of the examples you want to provide the minimum amount possible for the main examples. Its okay to break out into more complex cases to provide examples of overrides, but the basics should not need any extensions.

regimes/sa/org_identities.go
```diff
@@ -0,0 +1,27 @@
...
+
+func orgIdentityRules() *rules.Set {
+	return rules.For(new(org.Identity),
+		rules.When(
+			is.InContext(tax.RegimeIn(countryCode)),
+			rules.Field("code",
+				rules.Assert("01", "identity code must be valid",
```
samlown: Assertion could provide more details here, such as "SA identity code must be valid"

regimes/sa/sa.go
```diff
@@ -0,0 +1,66 @@
...
+		Description: i18n.String{
+			i18n.EN: here.Doc(`
+				Saudi Arabia's tax system is administered by the Zakat, Tax and
+				Customs Authority (ZATCA). VAT was introduced on January 1, 2018,
+				at 5% and increased to 15% on July 1, 2020.
+
+				Businesses must register for VAT if annual taxable supplies exceed
+				SAR 375,000, with voluntary registration available above SAR 187,500.
+				Registered businesses receive a VAT Identification Number which must
+				appear on all tax invoices.
```
samlown: I'd try to avoid any temporal information in the descriptions, such as historical rates and registration profiles. I know some descriptions do have this, but it should be avoided.

addons/sa/zatca/scenarios.go
```diff
@@ -0,0 +1,24 @@
...
+		Schema: bill.ShortSchemaInvoice,
+		List: []*tax.Scenario{
+			{
+				Types: []cbc.Key{
+					bill.InvoiceTypeStandard,
+				},
+				Ext: tax.ExtensionsOf(cbc.CodeMap{
+					untdid.ExtKeyDocumentType: "388",
```
samlown: Is this really required? Won't the en16931 addon handle this?

samlown (changes requested):
In addition to comments, we need to check if there is a way to normalize the invoices types into the ZATCA codes automatically, at least for the main cases. Exception cases can be handled with the "`other`" type.

addons/sa/zatca/scenarios.go
migueltorresvalls: This untdid document type is ZATCA specific and differs from the EN16931 set (maps to 380). The remaining of them are the same.

addons/sa/zatca/bill.go
migueltorresvalls: Wouldn't calculate set it to current time? 
 https://github.com/invopop/gobl/blob/ff019e9cb71ae08b9636c0c06050d6886c97b8a1/bill/calculator.go#L228-L232

examples/sa/simplified-zero-rated-with-notes.yaml
migueltorresvalls: We can create scenarios for the most used cases and leave more specific ones to be introduced by the customer along with the `InvoiceTypeOther`. Adding support for all transaction codes would require 32 scenarios

CHANGELOG.md
```diff
@@ -13,13 +13,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
...
 
 ### Added
 
 - `tax`: Added `CorrectionNormalize` callback type and `Normalize` field on `CorrectionDefinition` for addon-specific correction logic.
 - `bill`: Added `CorrectionOptionsValue()` accessor on `Invoice` for use by correction normalizers.
 - `cbc.CodeMap`: added `Lookup` method that returns the code matching a given key, falling back hierarchically to less specific keys.
 - `pay`: added `MeansKeyCredit` and `MeansKeyDebit` qualifiers, enabling the `card+credit` and `card+debit` payment means. Adapted all addons mapping payment means to extensions to use the two new qualified means.
+- `sa`: Added Saudi Arabia tax regime, including tax categories, tax identity validation, and organization identity types.
+- `sa-zatca-v1`: Added ZATCA addon for Saudi Arabia Phase 2 e-invoicing, covering standard and simplified invoices, credit and debit notes, party and line-level validation rules, tax-combo rules with VATEX support, invoice scenarios, and the `zatca-qr` stamp key for the base64-encoded TLV QR carried by every ZATCA invoice.
```
cavalle: @migueltorresvalls I think the last merge of main probably messed up the CHANGELOG. The new entries should be in the `Unreleased` section, not in the already released section

migueltorresvalls: Thank you for the catch! This is now fixed

addons/sa/zatca/bill.go
samlown: Okay, leave a comment then.

examples/sa/simplified-zero-rated-with-notes.yaml
samlown: Surely the default can be calculated?

regimes/sa/tax_identity.go
```diff
@@ -0,0 +1,29 @@
...
+// Saudi VAT Identification Number.
+func normalizeTaxIdentity(tID *tax.Identity) {
+	tax.NormalizeIdentity(tID)
+}
+
+func taxIdentityRules() *rules.Set {
+	return rules.For(new(tax.Identity),
+		rules.When(tax.IdentityIn(CountryCode),
+			rules.Field("code",
+				rules.Assert("01", "VAT number must be 15 digits starting/ending with 3",
```
samlown: Copy here should identify the object:

```suggestion
				rules.Assert("01", "tax identity code for SA must be 15 digits long, and start and end with 3",
```

You may also just want to consider two rules.

examples/sa/simplified-zero-rated-with-notes.yaml
migueltorresvalls: Yes, this has been moved to: https://github.com/invopop/gobl.sa.zatca/pull/1

bill/invoice_tags.go
```diff
@@ -127,6 +127,17 @@ var defaultInvoiceTags = &tax.TagSet{
...
+			Key: tax.TagExport,
+			Name: i18n.String{
+				i18n.EN: "Export",
+			},
+			Desc: i18n.String{
+				i18n.EN: here.Doc(`
+					Indicates that the invoice refers to an export of goods.
+				`),
+			},
+		},
```
samlown: This tag feels incoherent with the rest and is normally implied at the line level. Normally these tags are used to add notes or additional validations, which I cannot see in this PR.

addons/sa/zatca/bill.go
migueltorresvalls: Also moved to: https://github.com/invopop/gobl.sa.zatca/pull/1

samlown (changes requested):
I'm not sure the "export" tag is the best approach.

bill/invoice_tags.go
migueltorresvalls: I added it because it was already defined as a constant but not used in defaultInvoiceTags. Its used in the addon's scenarios here: https://github.com/invopop/gobl.sa.zatca/pull/1/changes#diff-a2ef87da2c4c496923feefb75bc17e0248e375a516f88e5b27de78d7b3875b8d

regimes/sa/tax_identity.go
migueltorresvalls: Two rules would imply having two regex for what could be done with one. For me it makes the code harder to read (but validation rules easier to understand). Leaving this up to you

bill/invoice_tags.go
migueltorresvalls: Moved to gobl.sa.zatca

samlown (approved):
Nice one!

regimes/sa/tax_identity.go
samlown: its more about the clarity for debugging, the regexp knows which is incorrect, but the user might not.

PR #787, opened 2026-04-07
Fixing extracting rules from schema.Object

alvarolivie:
- Replace this bullet list with your changes

## Pre-Review Checklist

- [ ] Opened this PR as a draft
- [ ] Read the CONTRIBUTING.md guide.
- [ ] Performed a self-review of my code.
- [ ] Added thorough tests with at **least** 90% code coverage.
- [ ] Modified or created example GOBL documents to show my changes in use, if appropriate.
- [ ] Added links to the source of the changes in tax regimes or addons, either structured or in the comments.
- [ ] Run `go generate .` to ensure that the Schemas and Regime data are up to date.
- [ ] Reviewed and fixed all linter warnings.
- [ ] Been obsessive with pointer nil checks to avoid panics.
- [ ] Updated the CHANGELOG.md with an overview of my changes.
- [ ] Marked this PR as ready for review.

And if you are part of the org:
- [ ] Requested a review from Copilot and fixed or dismissed (with a reason) all the feedback raised.
- [ ] Requested a review from @samlown.

PR #788, opened 2026-04-07
Fix rules context collection for Embeddable types

invopop-dev:
## Summary
- `collectContext` in `rules/context.go` did not recurse into `Embeddable` types (e.g. `schema.Object`), so regime/addon context was never collected when validating an `Envelope`. This caused context-guarded rules (FR, DE, CO, etc.) to silently skip during envelope validation.
- Fixed CO DIAN `invoiceNotSimplified` guard which was incorrectly scoped inside `Field("customer")` — it received `*org.Party` instead of `*bill.Invoice`, making it always return true (never simplified).
- Updated DE and FR examples that were exposed as invalid by the fix.

## Test plan
- [x] `TestEnvelopeValidate/with_more_complex_document_and_rules` now passes
- [x] Full test suite passes (`go test ./...`)
- [x] Example golden files regenerated

### DIFF
--- a/addons/co/dian/bill_invoices.go
+++ b/addons/co/dian/bill_invoices.go
@@ -86,16 +86,18 @@ func billInvoiceRules() *rules.Set {
 		),
 		// Customer validation
 		// Code 05: customer tax_id required when not simplified
-		// Code 06: Colombian customer must have at least one address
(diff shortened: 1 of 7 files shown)


samlown:
Replaces #787

PR #791, opened 2026-04-08
Refactor error handling: replace HTTP codes with domain error keys

invopop-dev:
## Summary

- Removes `cli.Error` type entirely — all error handling now uses `gobl.Error` directly
- Adds `ErrInput` and `ErrNotFound` sentinel errors to `gobl.Error` for input validation and missing resource cases
- Removes `ErrUnmarshal` in favour of `ErrInput`
- HTTP status codes are no longer embedded in errors — derived solely in the API layer's `writeError` via key-to-status mapping (`input`→400, `not-found`→404, `internal`→500, all others→422)
- API handlers use `gobl.ErrInput.WithReason(...)` / `gobl.ErrNotFound.WithReason(...)` instead of constructing `cli.Error` structs
- MCP layer updated to check for `*gobl.Error` instead of `*cli.Error`

## Test plan

- [x] `go test ./...` passes
- [x] API tests verify correct HTTP status codes are still returned
- [x] CLI error string format updated from `code=N, message=X` to `key: message`
- [x] Bulk response JSON no longer includes `"code"` field, uses `"key"` instead


schema/object.go
```diff
@@ -75,7 +75,7 @@ func objectRules() *rules.Set {
 // and return them as a list of faults. This will only check the
 // payload of the object, which would not otherwise be verified.
 func (d *Object) Validate() rules.Faults {
-	return rules.Validate(d)
+	return rules.Validate(d.Instance())
```
alvarolivie, line 78: Just want to make sure this was on purpose. Makes sense but seems out of place in this PR.

alvarolivie (approved):
Looks good! Just wondering if we could use these errors in apps as well.

schema/object.go
samlown: Claude might have missed commenting on this... it was intentional

PR #808, opened 2026-04-15
Remove customer tax id required for Poland

mrdanwa:
After reviewing the KSeF FA(3) documentation, third-country buyers and consumers don't need a tax ID on the invoice.

Check https://github.com/invopop/gobl.ksef/pull/56

## Pre-Review Checklist

- [ ] Modified or created example GOBL documents to show my changes in use, if appropriate.
- [ ] Added links to the source of the changes in tax regimes or addons, either structured or in the comments.
- [ ] Reviewed and fixed all linter warnings.
- [ ] Updated the CHANGELOG.md with an overview of my changes.


addons/pl/favat/bill.go
```diff
@@ -72,9 +72,6 @@ func billInvoiceRules() *rules.Set {
 		rules.When(is.Func("not simplified", invoiceNotSimplified),
 			rules.Field("customer",
 				rules.Assert("09", "customer is required", is.Present),
-				rules.Field("tax_id",
```
pmenendz, line 75: Is it always not required or only for specific cases? For examples foreign customer or simplified invoices

pmenendz (changes requested):

addons/pl/favat/bill.go
mrdanwa: I added some comments in the PR for gobl.ksef, it seems that for all regular invoices to consumers it does not need tax id, and there is no other field to know if business or consumer. I tried sending to to favat without tax id for different countries and it all worked.

pmenendz: Ok, then the validation would be that if tax ID and taxID.Country is PL, then the code is mandatory.

mrdanwa: Makes sense

addons/pl/favat/bill.go
```diff
@@ -121,6 +122,19 @@ func invoiceNotSimplified(val any) bool {
 	return !inv.HasTags(tax.TagSimplified)
 }
 
+// invoiceCustomerPLTaxIDCodePresent returns false when the customer has a
+// Polish tax ID but no code — the NIP is mandatory for Polish entities.
+func invoiceCustomerPLTaxIDCodePresent(val any) bool {
```
pmenendz: This can be probably simplified using rules.When. I am not that familiar with rules yet, so I'll let @samlown take a look.

The logic is fine though 👍

pmenendz (dismissed):

samlown (approved):
LGTM!

PR #814, opened 2026-04-22
Drop addon version from rule fault codes

samlown:
## Summary

- Fault codes for addon validation rules no longer embed the addon version. `GOBL-AR-ARCA-V4-BILL-INVOICE-24` is now `GOBL-AR-ARCA-BILL-INVOICE-24`. Generated files in `data/rules/` lose the version from their filenames too (`ar-arca.json`, not `ar-arca-v4.json`).
- Each addon package now exposes a `Key` constant for the unversioned family (e.g. `ar-arca`); versioned keys such as `V4` are derived via constant concatenation (`Key + "-v4"`). The `tax.AddonIn(...)` guard still uses the versioned key, so rules fire only when that specific addon version is enabled.
- Assertion IDs are now a cross-version contract: preserved rules must keep their existing numeric IDs on a version bump, and an ID must never be reassigned to a semantically different rule. Documented on `rules.Register`.

## Motivation

Most rules carry over unchanged when an addon bumps (v4 → v5, v3 → v4). Embedding the version in the fault code forced every customer error handler that pins a code to be rewritten on every bump, even when nothing about the rule actually changed. Dropping the version makes the fault-code namespace match the rule's actual lifetime.

## Breaking change

Consumers pinning fault-code strings should drop the version segment. A one-line regex handles the common case:

```
s/GOBL-([A-Z0-9]+-[A-Z0-9]+)-V[0-9]+(-)/GOBL-\1\2/
```

The published rule-doc pages in `gobl.docs` are regenerated from the JSON files in this repo, so they'll update on the next docs sync.

## Test plan

- [x] `go test ./...` passes (including `rules/codes_test.go` uniqueness/format checks).
- [x] `go run ./rules/generate.go` produces clean unversioned filenames and codes.
- [x] Spot-check one regenerated file to confirm codes inside lost the version segment.
- [ ] Verify downstream modules (`gobl.verifactu`, `gobl.ticketbai`, `gobl.cfdi`, etc.) — grep for pinned code strings and coordinate version bumps if needed.

### DIFF
--- a/data/rules/currency.json
+++ b/data/rules/currency.json
@@ -3,6 +3,22 @@
   "package": "currency",
   "subsets": [
     {
+      "id": "GOBL-CURRENCY-CODE",
+      "object": "currency.Code",
+      "subsets": [
+        {
+          "guard": "present",
+          "assert": [
+            {
+              "id": "GOBL-CURRENCY-CODE-01",
+              "desc": "currency code must be defined in GOBL",
+              "tests": "valid currency code"
+            }
+          ]
+        }
+      ]
+    },
+    {
       "id": "GOBL-CURRENCY-AMOUNT",
       "object": "currency.Amount",
       "subsets": [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Changed
 
 - `bill`: removed the default exchange-rate / regime-currency conversion check; addons now enforce currency convertibility where required.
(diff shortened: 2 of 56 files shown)


data/rules/currency.json
```diff
@@ -2,6 +2,22 @@
   "id": "GOBL-CURRENCY",
   "package": "currency",
   "subsets": [
+    {
```
alvarolivie, line 5: Double check this is not from the currency PR which did not get generated ast time?

alvarolivie (approved):
LGTM

cavalle (approved):
Very sensible change!

PR #818, opened 2026-04-23
Refactor tax.Extensions to immutable struct and sort cbc.Meta JSON keys

samlown:
## Summary

- `tax.Extensions` becomes an immutable struct wrapping an unexported `map[cbc.Key]cbc.Code`. JSON output is deterministic (alphabetical key order) via a custom `MarshalJSON`. All mutation methods clone and return a new instance, eliminating the "Set on nil gets silently lost" footgun. New helpers: `MakeExtensions`, `ExtensionsOf`, `ExtMap` alias, `Clone`, `Keys`, `All` (iter.Seq2), `IsZero`, `Len`, `ExtensionsFromValue`. Field tags migrated from `json:"ext,omitempty"` to `json:"ext,omitzero"` (Go 1.24).
- `cbc.Meta` keeps its `map[cbc.Key]string` type but gains deterministic JSON output and `Keys()`, `Values()`, `All()` iterator methods. All existing call sites work unchanged.
- Six latent bugs where `x.Ext.Set(...)` was called without assigning the result (e.g. `addons/gr/mydata/bill_charges.go`, `addons/fr/ctc/org.go`) are fixed as part of the Extensions migration.

## Breaking changes (Go consumers)

- `tax.Extensions{"key": "value"}` literals no longer compile. Use `tax.ExtensionsOf(tax.ExtMap{"key": "value"})` for table-style constructions or `tax.MakeExtensions().Set(...)` for fluent chains.
- `em[k]` / `em[k] = v` / `range em` / `len(em)` / `make(tax.Extensions)` patterns must use the new method API (`Get`, `Set`, `All`, `Len`, `MakeExtensions`).
- Validation guards using `val.(tax.Extensions)` should switch to `tax.ExtensionsFromValue(val)` to accept both value and pointer forms passed by the rules framework.
- `tax.CleanExtensions(em)` free function is removed; callers use `em.Clean()`.
- JSON shape is **unchanged** — downstream consumers parsing GOBL JSON are not affected.

## Test plan

- [x] `go build ./...` clean.
- [x] `go vet ./...` clean.
- [x] `go test ./...` — all 72 packages pass.
- [x] `tax/extensions.go` test coverage 93.8%; every function at 89% or above.
- [x] `cbc/meta.go` test coverage at 100% for new methods; `MarshalJSON` at 89.5% (uncovered lines are unreachable string-marshal error returns).
- [ ] Downstream sibling repos (`gobl.facturae`, `gobl.cfdi`, etc.) consume GOBL via `bill.*` / `org.*` types, which still embed Extensions / Meta by value and emit the same JSON shape. A coordinated bump will be needed when this ships, but no API changes are required in those repos.


cavalle (approved):
LGTM

PR #827, opened 2026-04-30
Refactor pay.Advance into pay.Record and support multiple methods on bill.Payment

samlown:
## Summary

- Lifts the shape that was `pay.Advance` into a canonical `pay.Record` (single payment event: means + amount + currency + optional date/percent/extensions) and reuses it both for advances on invoices and for the new `bill.Payment.Methods` slice.
- Replaces `bill.Payment.Method *pay.Instructions` with `bill.Payment.Methods []*pay.Record` so a payment document can record multiple methods, potentially in different currencies (required for Portugal's SAF-T, among others).
- Removes the `pay.Record.Grant` boolean; FacturaE consumers should set the new `es-facturae-subsidy` extension instead.

### `pay`
- **Breaking**: `pay.Advance` renamed to `pay.Record`. The transitional `pay.Method` and the original `pay.Advance` names are removed — no aliases. Schema URL is now `https://gobl.org/draft-0/pay/record` (the previous `pay/advance` schema is gone).
- New fields on `pay.Record`: `DirectDebit *DirectDebit`, `Online []*Online` so it can fully describe any payment means recorded on a payment document.
- `Grant bool` removed.
- `description` is no longer required by validation (it was a poor fit for payment-document methods); contexts that need it should enforce it themselves.
- `(*Record).Normalize` now accepts `tax.Normalizers` and threads them through to the means-specific nested structs (`Card`, `CreditTransfer`, `DirectDebit`, `Online`) so addons can extend them directly.

### `bill`
- **Breaking**: `bill.Payment.Method` (`*pay.Instructions`) replaced by `bill.Payment.Methods` (`[]*pay.Record`).
- Documents using the legacy singular `method` JSON field migrate transparently via `UnmarshalJSON` into a single-element `methods` array.
- When exactly one method is present and its `amount` is zero, `Calculate()` auto-fills it from the document `Total`.
- New validation rule (`GOBL-BILL-PAYMENT-08`): the sum of method amounts (converted to the document currency via `ExchangeRates` where needed) must equal `Total`. Partial payments must have method amounts that line up with the line totals being paid.

### `addons/es/facturae`
- New `es-facturae-subsidy` extension (values `S`/`N`, matching the Spanish convention used by `es/sii` and `es/verifactu`) for flagging advance payments that come from a public grant or subsidy. Replaces the removed `pay.Record.Grant` boolean.

### Cascade
- `[]*pay.Advance` → `[]*pay.Record` and `*pay.Advance` → `*pay.Record` across `bill/`, `regimes/pt/`, `tax/`, and every addon (`br/nfe`, `gr/mydata`, `it/sdi`, `mx/cfdi`, `pl/favat`, `pt/saft`).
- Renamed `normalizePayAdvance` → `normalizePayRecord` in those addons for consistency.
- Schemas, examples, and rules data regenerated.

## Test plan

- [x] `go test ./...` — clean
- [x] `go vet ./...` — clean
- [x] Coverage: `bill` 94.9%, `pay` 94.5% (well above the 90% target)
- [x] Existing PT/ES payment example fixtures still round-trip
- [x] New ES partial-payment example (`examples/es/payment-partial.yaml`) demonstrates a multi-currency, multi-line, multi-method payment with exchange rates
- [x] `payment-old.yaml` exercises the legacy `method` JSON migration path
- [ ] Smoke-test the change against `gobl.facturae` to confirm the `es-facturae-subsidy` extension lands where the old `Grant` boolean was being read


data/rules/gr-mydata.json
```diff
@@ -378,14 +378,14 @@
...
-      "object": "pay.Advance",
+      "id": "GOBL-GR-MYDATA-PAY-RECORD",
+      "object": "pay.Record",
       "subsets": [
         {
           "field": "key",
           "assert": [
             {
-              "id": "GOBL-GR-MYDATA-PAY-ADVANCE-01",
+              "id": "GOBL-GR-MYDATA-PAY-RECORD-01",
```
cavalle, line 388: I just realised that the rename in error codes are backwards incompatible and could potentially break existing integrations. I would be very suprised, though, if we already have integrations relying on the error codes affected by this change, so I guess it's fine to do it, given we have a good reason for it. I'm just highlighting in case we didn't consider the implications.

pay/record.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/currency"
+	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/rules"
+	"github.com/invopop/gobl/tax"
+	"github.com/invopop/gobl/uuid"
+	"github.com/invopop/jsonschema"
+)
+
+// Record represents a single payment that has been or will be made via a
```
cavalle: "a single payment" in the description might be confusing/misleading specially in the context of a `bill.Payment` document. I'd suggest re-wording as "a single transaction".

cavalle (approved):
LGTM! I just left a couple of comments with some non-blocking input to be considered.

data/rules/gr-mydata.json
samlown: I'd forgotten about that also... as you say, very unlikely this will be an issue for anyone, but its something we'll have to take into account for the future.

PR #834, opened 2026-05-13
Add es-tbai-regime and es-tbai-identity-type extensions

pmenendz:
Adds two new TicketBAI extensions plus the customer-identity plumbing that goes with the second one.

## `es-tbai-regime`

Exposes `ClaveRegimenIvaOpTrascendencia` (L9 — codes 01-17, 51-53) so callers can set the regime explicitly. The Verifactu addon already exposes the equivalent `es-verifactu-regime`; the gap on the TicketBAI side meant integrators (and the `gobl.ticketbai` mapper) had no way to express anything other than the four hard-coded values the mapper inferred.

Normalization keeps the previous defaults (export → 02, equivalence surcharge → 51, `simplified-scheme` tag → 52, otherwise 01) and preserves explicit values. Validation will land in a follow-up.

## `es-tbai-identity-type`

Exposes the L7 `IDType` code (subset: `02` NIF-VAT/VIES, `03` passport, `04` foreign ID, `05` resident certificate, `06` other) for use under `IDOtro` when there is no usable tax ID for the customer.

`normalizeInvoicePartyIdentity` maps the customer's first identity to the extension based on its `key` (`passport` → `03`, `foreign` → `04`, `resident` → `05`, `other` → `06`). Spanish NIFs and EU/non-EU tax IDs continue to be handled via the regular tax-identity path and need no extension.

## Customer tax_id rule relaxed

Rule **08** previously required `customer.tax_id` on non-simplified invoices. It now accepts either a tax ID **or** an identity carrying the `es-tbai-identity-type` extension — mirroring the existing Verifactu behaviour. Identities without that extension still fail validation.

## Pre-Review Checklist


addons/es/tbai/bill.go
```diff
@@ -28,6 +28,97 @@ func normalizeInvoice(inv *bill.Invoice) {
...
+		fallback = "52"
+	}
+	forEachVATCombo(inv, func(tc *tax.Combo) {
+		tc.Ext = tc.Ext.SetIfEmpty(ExtKeyRegime, fallback)
+	})
+}
+
+// forEachVATCombo applies fn to every VAT/IGIC/IPSI combo in the invoice's
+// lines, charges, and discounts. Nil entries are skipped.
+func forEachVATCombo(inv *bill.Invoice, fn func(*tax.Combo)) {
```
pmenendz: This should normally be handled in tax combo normalizer, but Ticketbai uses some legacy tags like simplified-scheme to define the regime extension

addons/es/tbai/bill.go
```diff
@@ -28,6 +28,55 @@ func normalizeInvoice(inv *bill.Invoice) {
...
 	}
 	normalizeInvoiceTax(inv)
+	normalizeInvoiceSimplified(inv)
+	normalizeInvoicePartyIdentity(inv.Customer)
+}
+
+// normalizeInvoiceSimplified sets es-tbai-simplified = ~S~ on the invoice's
+// tax extensions when the generic ~simplified~ tag is present, so downstream
+// conversion can route on the extension rather than the tag. The default
+// (no tag) is left unset and treated as ~N~ by consumers.
```
cavalle: I don't like this type of Claude comments that reveal implementation details. They're not consistent we how we typically document code.

addons/es/tbai/extensions.go
```diff
@@ -8,11 +8,29 @@ import (
...
-	ExtKeyProduct    cbc.Key = "es-tbai-product"
-	ExtKeyCorrection cbc.Key = "es-tbai-correction"
-	ExtKeyBIActivity cbc.Key = "es-tbai-bi-activity"
+	ExtKeyRegion       cbc.Key = "es-tbai-region"
+	ExtKeyExempt       cbc.Key = "es-tbai-exemption"
+	ExtKeyProduct      cbc.Key = "es-tbai-product"
+	ExtKeyCorrection   cbc.Key = "es-tbai-correction"
+	ExtKeyBIActivity   cbc.Key = "es-tbai-bi-activity"
+	ExtKeyRegime       cbc.Key = "es-tbai-regime"
+	ExtKeyIdentityType cbc.Key = "es-tbai-identity-type"
```
cavalle, line 17: `es-tbai-identity-type` may be a bit long. But maybe is consistent with what we're doing in other addons, in which case I'd dismiss this comment. Alternatively, I think just `es-tbai-identity` would work.

cavalle (approved):
LGTM! Just a couple of nitpicks

addons/es/tbai/extensions.go
pmenendz: It is the same as verifactu

cavalle: Resolving then

addons/es/tbai/bill.go
```diff
@@ -94,15 +137,13 @@ func billInvoiceRules() *rules.Set {
...
-			is.Func("non-simplified", func(val any) bool {
-				inv, ok := val.(*bill.Invoice)
-				return ok && inv != nil && !inv.HasTags(tax.TagSimplified)
-			}),
+			is.Func("non-simplified", invoiceNotSimplified),
 			rules.Field("customer",
 				rules.Assert("03", "customer is required for non-simplified invoices", is.Present),
-				rules.Field("tax_id",
-					rules.Assert("08", "customer tax ID is required", is.Present),
+				rules.Assert("08", "customer must have a tax_id or an identity with ext 'es-tbai-identity-type'",
```
samlown: lacking context here, should say "invoice customer..."

addons/es/tbai/tax_combo.go
```diff
@@ -32,7 +45,7 @@ func normalizeTaxCombo(tc *tax.Combo) {
 				Set(ExtKeyExempt, "S2")
 		case tax.KeyOutsideScope:
 			tc.Ext = tc.Ext.
-				SetOneOf(ExtKeyExempt, "OT", "RL", "VT", "IE")
+				SetOneOf(ExtKeyExempt, "RL", "IE", "OT", "VT")
```
samlown, line 51: This implies "RL" is the default, is that expected? Reading further down there is a test for this, so maybe so. Probably worth a comment somewhere saying why its the default.

addons/es/tbai/tax_combo.go
```diff
@@ -43,6 +56,9 @@ func normalizeTaxCombo(tc *tax.Combo) {
 			tc.Ext = tc.Ext.
 				Set(ExtKeyExempt, "E5")
 		}
+	case es.TaxCategoryIGIC:
+		tc.Ext = tc.Ext.
+			Set(ExtKeyExempt, "IE")
```
samlown, line 67: Is this required?

PR #839, opened 2026-05-26
feat: ad regime

addaia:
## Pre-Review Checklist
- [x] Opened this PR as ready for review (not draft — external contributor)
- [ ] Added thorough tests with at least 90% code coverage (not measured,
      but all public functions and both valid/invalid paths are covered)
- [x] Modified or created example GOBL documents to show my changes in use
      (examples/ad/invoice.json)
- [x] Added links to the source of the changes in tax regimes or addons,
      either structured or in the comments.
- [x] Run `go generate .` to ensure that the Schemas and Regime data are
      up to date (data/regimes/ad.json generated and committed)
- [x] Reviewed and fixed all linter warnings

Not part of the org, skipping org-specific review steps.


regimes/ad/ad.go
```diff
@@ -0,0 +1,95 @@
...
+				its own indirect tax regime.
+
+				The Impost General Indirecte (IGI), introduced on 1 January 2013 by
+				Llei 11/2012, replaced the previous sales tax system. IGI applies at
+				five rates: a super-reduced rate (0%) for healthcare, education, gold
+				investment and certain non-profit services; a reduced rate (1%) for
+				food, books and magazines; a special rate (2.5%) for public transport
+				and cultural services; the general rate (4.5%); and an increased rate
+				(9.5%) for banking and financial services. The 4.5% general rate is
+				the lowest standard indirect-tax rate in Europe.
```
samlown: We try to avoid including dynamic information inside the descriptions to avoid accidental differences.

regimes/ad/bill_invoices.go
```diff
@@ -0,0 +1,56 @@
...
+//   - an org identity of type NRT (for entities below the threshold who
+//     hold an NRT but are not required to register for IGI).
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(CountryCode)),
+			rules.Field("supplier",
+				rules.Assert("01",
+					"invoice supplier in Andorra must have an NRT tax ID code or an NRT identity",
+					is.Func("has NRT", hasSupplierNRT),
```
samlown: Its usually clearer to try and use the `is` related methods rather than breaking out into a function.

regimes/ad/tax_identity.go
```diff
@@ -0,0 +1,68 @@
...
+
+var reNRT = regexp.MustCompile(`^[A-Z]\d{6}[A-Z]$`) // used [A-Z] for the prefix rather than [FEACDGLOPU] 
+// deliberately. The OECD document doesn't claim its list is exhaustive
+
+func taxIdentityRules() *rules.Set {
+	return rules.For(new(tax.Identity),
+		rules.When(tax.IdentityIn(CountryCode),
+			rules.Field("code",
+				rules.AssertIfPresent("01", "invalid Andorran tax identity code (NRT)",
+					is.Func("valid NRT", isValidTaxIdentityCode),
```
samlown: I think this could have been simplified to just use the `is.Matches` method directly.

regimes/ad/org_identities.go
```diff
@@ -0,0 +1,95 @@
...
+		return
+	}
+	code := strings.ToUpper(id.Code.String())
+	code = tax.IdentityCodeBadCharsRegexp.ReplaceAllString(code, "")
+	code = strings.TrimPrefix(code, string(l10n.AD)) // strip "AD" country prefix
+	code = strings.TrimPrefix(code, "NRT")           // strip "NRT" label prefix
+	id.Code = cbc.Code(code)
+}
+
+func orgIdentityRules() *rules.Set {
```
samlown: Are you sure this check and definition is required? If the NRT is an official tax ID, it is probably only required in the tax.Identity.

regimes/ad/tax_categories.go
```diff
@@ -0,0 +1,159 @@
...
+// All five rates and their descriptions are sourced directly from the
+// Departament de Tributs i de Fronteres official page:
+// https://www.e-tramits.ad/tramits/impostos/igi
+//
+// IGI was introduced on 1 January 2013 by Llei 11/2012, del 21 de juny.
+// Rates have remained stable since introduction.
+func taxCategories() []*tax.CategoryDef {
+	return []*tax.CategoryDef{
+		{
+			Code: tax.CategoryVAT,
```
samlown: Are you sure `VAT` is the right category? Sounds like the code here should be IGI.

regimes/ad/ad.go
```diff
@@ -0,0 +1,95 @@
...
+		taxIdentityRules(),
+	)
+}
+
+// New provides the tax region definition for Andorra.
+func New() *tax.RegimeDef {
+	return &tax.RegimeDef{
+		Country:   CountryCode,
+		Currency:  currency.EUR,
+		TaxScheme: tax.CategoryVAT,
```
samlown: Why tax scheme VAT when the description specifically says IGI is the regional tax?

samlown (changes requested):
Thanks for this, but there are a few issues. The local main tax scheme does not appear to be VAT. Supporting the same tax ID in both org identities and tax identity is probably not required. There could be some simplification around the rules handling and usage of break-out methods. I suspect the tests haven't been run as expect the lack of a generated example to have raised a failure.

addaia (commented):
edited per comments

PR #842, opened 2026-05-28
Fix correction type duplication in PT

cavalle:
- `tax.CorrectionDefinition.Merge` now deduplicates merged types, extensions, and stamps, preventing duplicate entries when both a regime and an addon declare the same keys.
- Removed correction definition types from `pt-saft-v1` already defined in the regime.

## Pre-Review Checklist


pmenendz (approved):
LGTM 👍

addons/pt/saft/invoices.go
```diff
@@ -19,11 +19,7 @@ import (
 
 var corrections = tax.CorrectionSet{
 	{
-		Schema: bill.ShortSchemaInvoice,
-		Types: []cbc.Key{
-			bill.InvoiceTypeCreditNote,
-			bill.InvoiceTypeDebitNote,
-		},
```
samlown, line 26: Is this really the case? Corrections can be of any type?

samlown (approved):
Seems okay, I'd just if you really want to allow any type of correction document.

addons/pt/saft/invoices.go
cavalle: Not exactly. The same correction types were declared both in the addon and in the regime. In other countries, we only have them in the regime. I just kept the ones in the PT regime for consistency with what we do in the rest of the countries.

PR #843, opened 2026-05-29
fix(org): keep dotted inbox codes in code, not url

alvarolivie:
## Problem

A Peppol participant identifier whose code contains dots (e.g. `BE0425.260.668`) was being normalized into the inbox's `url` field instead of staying in `code`.

## Root cause

`org.Inbox.Normalize()` routed any code accepted by `is.URL` into `url`. `is.URL` is backed by the permissive `govalidator.IsURL`, which accepts bare domain-like strings with no scheme

## Fix

Only route a code into `url` when it carries an explicit scheme (`://`). This is consistent with the field's own doc ("URL of the inbox that includes the protocol, server, and path") and preserves the existing `https://inbox.example.com` behaviour, while leaving schemeless identifiers (dotted Peppol codes) in `code`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


alvarolivie:
This was a test PR

PR #845, opened 2026-06-01
Add New Zealand (NZ) GST regime

pedrorfdez:
### Adds a tax regime for New Zealand (NZ).

- GST category with standard rate (15%, since Oct 2010), historical rates (12.5% since Jul 1989, 10% since Oct 1986), and accommodation reduced rate (9%, applying to stays of 28 or more consecutive days, equivalent to 60% of the standard rate)
- Zero-rated and exempt supply categories via the global GST keys
- IRD number (Inland Revenue Department number) validation using a two-pass modulo-11 checksum; pass-2 fallback weights handle the edge case where pass-1 produces an impossible check digit of 10
- NZBN (New Zealand Business Number) as an optional org identity, validated with a GS1 Modulo-10 check digit
- Credit note and debit note correction types, mapping to NZ's Supply Correction Information (SCI)
- Example invoice demonstrating a standard-rated services line and a zero-rated export line

*Invoice-level threshold requirements (the $200/$1,000 Taxable Supply Information rules) and Peppol/PINT A-NZ e-invoicing format support are intentionally deferred to a future addon, following the precedent set by the Singapore (SG) and Uruguay (UY) base regime implementations.*

Sources:
- IRD number validation: https://www.ird.govt.nz/digital-service-providers/services-catalogue/customer-and-account/ird-number-validation
- OECD NZ TIN documentation: https://www.oecd.org/content/dam/oecd/en/topics/policy-issue-focus/aeoi/new-zealand-tin.pdf
- GST rates and history: https://www.ird.govt.nz/gst
- NZBN specification: https://www.nzbn.govt.nz/whats-an-nzbn/about/
- GS1 GLN check digit: https://www.gs1nz.org/services/glns/

## Pre-Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Added
+
+- `regimes/nz`: New Zealand (NZ) GST regime. Includes standard rate (15%, with historical 12.5% and 10% rates), accommodation reduced rate (9%), zero-rated and exempt supplies, IRD number validation with two-pass modulo-11 checksum, NZBN (New Zealand Business Number) as an optional organisation identity with GS1 Mod-10 checksum, and credit/debit note corrections.
+
 ## [v0.501.0] - 2026-06-16
 
 ### Added
--- a/data/regimes/nz.json
+++ b/data/regimes/nz.json
@@ -147,6 +147,24 @@
               "percent": "6%"
             }
           ]
+        },
+        {
+          "rate": "zero",
+          "keys": [
+            "standard"
+          ],
+          "name": {
+            "en": "Zero Rate"
+          },
+          "desc": {
+            "en": "Zero-rated supplies including exported goods, international transport, and certain land transactions between GST-registered parties."
+          },
+          "values": [
+            {
+              "since": "1986-10-01",
+              "percent": "0%"
+            }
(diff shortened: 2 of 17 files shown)


pedrorfdez:
I added `KeyReverseCharge` to `GlobalGSTKeys()` in `tax/constants.go` as part of the NZ reverse-charge scenario. This mirrors what the VAT global keys already have. This is a cross-regime change that would also benefit CA, SG, and IN. Happy to revert it to NZ-local if you'd prefer to keep `GlobalGSTKeys` minimal for now.

Also replaced the initial `bill.InvoiceScenarios()` in the `Scenarios` field with a NZ-specific scenario following `regimes/in`. `InvoiceScenarios()` gates on `Categories: [CategoryVAT]` so it can never fire for GST-only documents. I'd be happy to track this in an issue to implement a proper generalisation (if that makes sense, maybe it is designed like this on purpose).

pedrorfdez:
I've also found a bug that does not allow Windows based systems to run the `TestConvertExamplesToJSON` test due to the path containing backslashes instead of usual slashes: e.g.: `.\data\data.go` instead of `./data/data.go`, so when comparing with `skipExamplePaths` defined paths, it fails. Will submit a different PR to fix this, solution is a one line change, replacing `path` by `filepath.ToSlash(path)`.

Edit: I was working on an old main version, looks like the issue is still there, just in a different place. Fix in #867

PR #846, opened 2026-06-02
Fix int64 overflow in Amount.Multiply for prices with many decimal places

### LINKED ISSUE #844: bug: integer overflow handling in line items
An issue has been discovered when dealing with prices with many decimal places and a large quantity. The current overflow protection is failing in this case. Example:

```
{
  "lines": [
    {
            "quantity": "439",
            "item": {
                "name": "Consumers",
                "description": "Price per consumer (volume based)",
                "price": "3.0888382687927107"
            }
    }
  ]
}
```

generates a line sum of `-922.34`.

pedrorfdez:
Fixes a silent `int64` overflow in `Amount.Multiply` that produced large negative line sums when a price with many decimal places was multiplied by a large integer quantity.

Fixes #844 

**Reproducer from the original issue:**
```json
{
  "lines": [{
    "quantity": "439",
    "item": { "price": "3.0888382687927107" }
  }]
}
```
Generated a line sum of -922.34 instead of 1356.00

**Root cause:**
Multiply used float64 as the intermediate representation. The price 3.0888382687927107 is stored internally as
Amount{value: 30888382687927107, exp: 16}. Multiplying by 439 produces ≈ 1.356 × 10¹⁹, which exceeds MaxInt64 ≈ 9.22 × 10¹⁸. On x86, int64(overflowing_float64) wraps to MinInt64, which at exp=16 prints as approximately -922.34. The existing AmmountMaxDigits guard only protects string parsing, not arithmetic.

**Fix:**
Replace float64 multiplication in Multiply with math/big integer arithmetic, eliminating both the int64 overflow and the float64 precision loss. If the result still cannot fit in int64 (e.g. two large integers with exp == 0), precision is reduced by one decimal place and the operation is retried.

**Tests added:**
The exact reproducer of the original issue

## Pre-Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Fixed
+
+- `num`: `Amount.Multiply` now uses `math/big` for the integer product and division instead of `float64`, preventing a silent `int64` overflow that produced large negative results when multiplying a price with many decimal places (e.g. `3.0888382687927107`) by a large integer quantity.
+
 ## [v0.403.0] - 2026-05-13
 
 ### Changed
--- a/num/amount.go
+++ b/num/amount.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"math/big"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


samlown:
Thanks for this @pedrorfdez! We ended up taking a modified approach in #852, but this got the ball rolling :-)

pedrorfdez:
> Thanks for this @pedrorfdez! We ended up taking a modified approach in #852, but this got the ball rolling :-)

You're welcome! Happy to help 😄

PR #847, opened 2026-06-03
Extract fr/ctc to an external module; add approved external-addon registry

samlown:
Reverses the in-core French CTC direction (PR #820): instead of carrying the
CTC addon in core, this promotes the general-purpose features it relies on into
core and moves the addon itself to a standalone, opt-in module —
[`github.com/invopop/gobl.fr.ctc`](https://github.com/invopop/gobl.fr.ctc).

## Commits

1. **Bring CTC core features into core; drop in-core fr/ctc addon** — promotes the
   general features developed alongside CTC so the addon can live externally on a
   released core:
   - `bill`: document lifecycle — `bill.Status` (`StatusLine`/`Reason`/`Action`)
     and advice/receipt `bill.Payment` types, with closed-set status-line keys.
   - `envelope`/`head`: envelope-level fault ignoring by fully-qualified code.
   - `rules`: `is.OneOf` and `is.AnyOf` (renamed from `is.Or`), plus by-code fault
     dropping.
   - `catalogues/dgfip`: DGFiP catalogue of French billing-mode extensions.
   - `tax`: `Addons.AddAddons` helper; `en16931.NormalizeTaxCombo` made public.

   The pre-existing flat in-core `fr/ctc` addon is removed here (superseded by the
   external module); its data and examples go with it.

2. **Add approved external-addon registry; recognise fr-ctc keys** — so an external
   addon's key stays a valid `$addons` value without its implementation compiled
   into core:
   - `tax.ExternalAddon` / `RegisterApprovedAddon` / `ApprovedAddons`
   - `tax/addons_external_list.go`: the reviewed approved list (the four `fr-ctc-*`
     keys); editing it is the approval step.
   - `AddonList.JSONSchemaExtend` unions registered defs with approved stubs, so the
     addon-list schema lists `fr-ctc-*` again.

## The runtime gate stays strict

Being on the approved list is **recognition/governance only, never a runtime
bypass**. A document declaring an approved-but-unloaded addon still fails
`add-on must be registered`. Any service that processes French CTC documents must
import `github.com/invopop/gobl.fr.ctc`.

## Follow-ups

- Supersedes and should close #820.
- After this merges + a core tag is cut, `gobl.fr.ctc` drops its dev `replace` and
  pins the tag.


samlown:
The missing code coverage is a bit hard to solve, so I think it can be ignored for this case 🤔

bill/invoice.go
```diff
@@ -259,6 +259,11 @@ func (inv *Invoice) Empty() {
 	inv.Payment.ResetAdvances()
 }
 
+// maxAddonResolutionPasses caps the iteration in Calculate that picks
+// up addons appended by meta-addon normalizers during normalization.
+// In practice one extra pass is enough; the cap is a safety net.
+const maxAddonResolutionPasses = 4
```
cavalle, line 265: Agreed one pass should be enough, and in fact, more than one pass should probably be an antipattern. Should set the cap to one to ensure we don't abuse of this feature?

bill/invoice.go
```diff
@@ -268,7 +273,22 @@ func (inv *Invoice) Calculate() error {
...
+		}
+	}
 	inv.Normalize(tax.ExtractNormalizers(inv))
+	for pass := 0; pass < maxAddonResolutionPasses; pass++ {
+		newNorms := tax.ExtractNormalizersForNew(inv, seen)
+		if len(newNorms) == 0 {
+			break
+		}
+		inv.Normalize(newNorms)
+	}
```
cavalle, line 291: This code is replicated in all document types. I was thinking it would be great to find a way to avoid the code being duplicated. Also: `bill.Order` and `bill.Delivery` should also probably have this logic.

head/header.go
```diff
@@ -33,6 +33,22 @@ type Header struct {
 
 	// Any information that may be relevant to other humans about this envelope
 	Notes string `json:"notes,omitempty" jsonschema:"title=Notes"`
+
+	// Ignore lists fully-qualified validation fault codes to suppress when
+	// validating this envelope, e.g. "GOBL-EU-EN16931-ORG-ITEM-01". Intended
+	// for format-conversion cases where a specific, known fault is acceptable.
+	// Covered by the envelope signature when sealed. NOTE: any code may be
+	// listed, including structural envelope/header codes — use deliberately.
+	Ignore []rules.Code `json:"ignore,omitempty" jsonschema:"title=Ignore"`
```
cavalle, line 42: If I understand correcly, this will allow users to bypass any rule they want. This could potentially produce invalid outputs in the converters. This could be particularly problematic in cases where the converted output is not validated by any external source. For instance, in NO VERI*FACTU or SAF-T, the output is stored in the databse.

In summary, today, converters rely on the addon validations to have been enforced as a pre-condition for the output to be valid. Are we ok with allowing users to skip rules and as a consequence generating invalid conversions?

cavalle (approved):
Left a few comments. In summary:
* External vs in-core addons make a lot of sense to me!
* Ignore: allowing issuers to arbitrarily ignore rules could lead to invalid conversion outputs that may go undetected until months after issuance (e.g. in NO VERI*FACTU or SAF-T workflows).
* Meta-addons: they introduce the additional complexity of correctly handling multiple calculation and validation passes. I’m not convinced that this added internal complexity is justified by the benefits. A silo.modify step could potentially serve the same purpose.

PR #852, opened 2026-06-10
num: fix int64 overflow in Amount arithmetic

### LINKED ISSUE #844: bug: integer overflow handling in line items

samlown:
Fixes silent `int64` overflows in `num.Amount` arithmetic. `Multiply`, `Divide` and `Rescale` used `float64` intermediates that could overflow and wrap to large negative results when handling values with many significant digits. `Add`, `Subtract` and `Compare` had related exposure: unguarded `int64` sums, and a rescaled operand that could silently lose its scale near the limit.

Fixes #844

Alternative to #846, which only fixed `Multiply` — the other operations have the same bug (confirmed on `main`, see the issue comments for reproducers).

## Approach

All operations now compute intermediates with `math/big` and share a single overflow policy (`fitBigToInt64`):

- Results that cannot be held at the requested exponent drop decimal places (with rounding) to fit.
- Values whose integer part exceeds `int64` saturate at the boundary instead of wrapping.
- Division by zero returns a zero amount.

`Amount` keeps its `int64` storage and value semantics; `big.Int` is only used internally during calculation.

## Tests

Issue reproducers for all operations, accumulated-sum overflow, mixed-exponent add/compare near the limit, boundary saturation, and division by zero. 100% coverage on the changed functions; full suite and linter pass.

## Note for reviewers

`Rescale(exp)` on a value too large for the requested exponent now returns the highest exponent that fits rather than overflowing — a small contract change worth a look. The same applies to `Add`/`Subtract`/`Multiply`/`Divide` results at the boundary.


num/amount.go
```diff
@@ -416,6 +432,76 @@ func intPow(base int, exp uint32) int64 { // nolint:unparam
...
+
+// fitBigToInt64 returns v as an int64 paired with the exponent it should be
+// stored at. If v already fits, it is returned unchanged at the requested
+// exponent. Otherwise the least-significant decimal digits are dropped (with
+// rounding) and the exponent lowered one place per digit, preserving the
+// integer magnitude at the cost of decimal precision — the trade-off that
+// suits the "human manageable domain" Amount targets. If the integer part
+// alone still exceeds int64 once the exponent reaches zero, the value
+// saturates at the int64 boundary so a genuinely out-of-range result never
+// silently wraps to the wrong sign.
```
cavalle: I don't like these long comments describing the function implementation details as part of the function description that Claude tends to add. I believe those comments, if they are really necessary, should better be part of the implementation details where the code is not clear enough.

num/amount.go
```diff
@@ -147,24 +148,36 @@ func (a Amount) Sub(b Amount) Amount {
 	return a.Subtract(b)
 }
 
-// Multiply the amount by the provided amount.
+// Multiply the amount by the provided amount. The calculation is performed
+// using big.Int arithmetic so that neither the intermediate product nor the
+// rescaling can overflow int64 or lose precision the way float64 would. See
+// fitBigToInt64 for how out-of-range results are handled.
```
cavalle: I think the old comment was fine. If anything, I'd say something like it uses big.Int arithmetic to avoid overflows or lose precision.

num/amount.go
```diff
@@ -147,24 +148,36 @@ func (a Amount) Sub(b Amount) Amount {
...
 }
 
-// Divide the amount by the provided amount. Floating points are used for the actual division
-// and then round again to get an int. This prevents rounding errors, but if you want true division
-// with a base and a remainder, use the Split method.
+// Divide the amount by the provided amount. As with Multiply, big.Int
+// arithmetic is used to avoid the int64 overflow and float64 precision loss
+// that plain division would suffer. If you want true division with a base and
+// a remainder, use the Split method instead. Dividing by zero returns a zero
+// amount rather than panicking.
```
cavalle: Same as my previous feedback. This is claude's context and reasoning leaking to the method's description.

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 
+- `num`: `Amount.Multiply`, `Amount.Divide`, and `Amount.Rescale` now perform their intermediate arithmetic with `math/big` instead of `float64`, fixing silent `int64` overflows that produced large negative results (e.g. a price like `3.0888382687927107` multiplied by a large quantity yielded `-922.34` instead of `1356.00`) and the float64 precision loss on values with many significant digits. When a result cannot be held at the requested exponent, the least-significant decimal places are dropped (with rounding) to keep the integer magnitude; a value whose integer part alone exceeds `int64` saturates at the boundary rather than wrapping. Dividing by zero now returns a zero amount instead of an undefined value.
```
cavalle: Again. We probably don't need that level of detail about the changes in the CHANGELOG.

cavalle (approved):
Left some feedback. Nothing blocking.

num/amount.go
samlown: This is an internal method, so I'm less worried about a bit verbosity. Let's see if we can make it simpler though.

samlown:
@cavalle agreed on the comments. I've simplified and also added support for some of the other methods (Add, Subtract, etc.)

PR #853, opened 2026-06-10
Register dk oioubl addon + EXCISE tax category

CarlosMari:
## Summary

Prepares GOBL core to support the externalized Danish OIOUBL 2.1 addon
([`invopop/gobl.dk.oioubl`](https://github.com/invopop/gobl.dk.oioubl)). Three
related changes:

### 1. Approve the external `dk-oioubl` addon

Adds `dk-oioubl-v2-1` to the approved external-addon registry (`addons/external.go`)
so GOBL accepts documents that declare it. The addon itself lives in its own
module (mirroring `gobl.fr.ctc`); this is the core-side allow-list entry plus the
`data/schemas/tax/addon-list.json` regeneration.

### 2. Add the Danish excise duty tax category

Adds `EXCISE` (punktafgift) to the DK regime — a tax distinct from VAT, levied on
specific goods (alcohol, tobacco, energy, …), with per-product rates supplied per
line rather than predefined (modelled like the Spanish `IPSI` category). This
gives OIOUBL a home for the `taxcategoryid-1.1` `Excise` value as its own tax
category rather than overloading a VAT rate. The remaining three OIOUBL tax
categories (`StandardRated`, `ZeroRated`, `ReverseCharge`) map cleanly from the
existing GOBL VAT keys and need no regime change.

## Pre-Review Checklist


addons/external.go
```diff
@@ -66,4 +66,13 @@ func init() {
 		},
 		Module: "github.com/invopop/gobl.sa.zatca",
 	})
+
+	tax.RegisterApprovedAddon(&tax.ExternalAddon{
+		Key: "dk-oioubl-v2-1",
```
samlown: Key here is too specific, we usually just stick to the major version for the addon: `v2`.

regimes/dk/tax_categories.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/invopop/gobl/tax"
 )
 
+// TaxCategoryExcise is the Danish excise duty (punktafgift), a tax distinct from
+// VAT levied on specific goods (alcohol, tobacco, energy, …). Its rates vary per
+// product, so they are supplied per line rather than predefined.
+const TaxCategoryExcise cbc.Code = "EXCISE"
```
samlown: Is this an actual code? It sounds like these are just charges on the invoice rather than taxes.

samlown (changes requested):
Version details need fixing. I'd clarify if this is the best approach.  The `EXCISE` code sounds invented...

regimes/dk/tax_categories.go
CarlosMari: In oioubl its classified as a tax category (There is VAT (Moms) and Excise (For Alcohol/Car Registration/ Sugar Tax…), but I do agree that in gobl it may actually map better as a charge as VAT can be applied to this "charge".

addons/external.go
CarlosMari: Removed from key, kept in description as 2.1 is supposed to be the last OIOUBL version. Next nemhandel document type will be NemHandel BIS 4 (based on Peppol BIS 4).

samlown (approved):

PR #856, opened 2026-06-11
org: add LiquidationState and SoleShareholder to Registration

aroundthfur:
Two free-form string fields on org.Registration for the Italian FatturaPA IscrizioneREA block: LiquidationState maps to <StatoLiquidazione> (LS/LN) and SoleShareholder maps to <SocioUnico> (SU/SM). Both are normalized via cbc.NormalizeString, matching the existing free-form treatment of Office, Entry, and the other registration fields. No enum validation.

The org.Registration doc-comment invites country-specific extensions; these fields let gobl.fatturapa emit a structurally correct IscrizioneREA for companies in liquidation and sole-shareholder SRLs instead of hardcoding <StatoLiquidazione> to "LN" and omitting <SocioUnico>.

The it/hotel example demonstrates both fields in use.

## Pre-Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Added
+
+- `org`: added `LiquidationState` and `SoleShareholder` free-form fields to `org.Registration`, supporting the Italian FatturaPA `<StatoLiquidazione>` (`LS`/`LN`) and `<SocioUnico>` (`SU`/`SM`) elements of the `IscrizioneREA` block.
+
 ## [v0.500.0] - 2026-06-10
 
 GOBL is now a pure document library. The CLI, HTTP API, MCP server, and WASM build have moved to the [gobl.dev](https://github.com/invopop/gobl.dev) project, which composes this library with the complete set of addons. Install the CLI from its new home: `go install github.com/invopop/gobl.dev/cmd/gobl@latest`.
--- a/data/schemas/org/registration.json
+++ b/data/schemas/org/registration.json
@@ -57,6 +57,16 @@
         "other": {
           "type": "string",
           "title": "Other"
+        },
+        "liquidation_state": {
+          "type": "string",
+          "title": "Liquidation State",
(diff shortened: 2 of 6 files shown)


aroundthfur:
@samlown when you have a moment, could you take a look at this one? It's a small addition (two free-form registration fields following the existing pattern) and invopop/gobl.fatturapa#68 is blocked on it landing in a gobl release. Thanks!

samlown (changes requested):
Many thanks for this contribution! I think these fields however are too region specific to be included in the general "Registration" object. I think what we need to do here instead is include an `ext` field inside the Registration and define two specific extensions for the fatturapa addon: `it-fatturapa-liquidation-state` and `it-fatturapa-shareholder-state`. I'm copying in my colleague @rilla who has recently been working on the IT integration.

aroundthfur:
Thanks @samlown, that makes sense, and I see #863 already landed the generic `Ext` field on `org.Registration`. I'll close out the two-field approach here in favour of that.

One thing to confirm before I add the actual extension definitions: in your review you named them `it-fatturapa-liquidation-state` and `it-fatturapa-shareholder-state`, but the existing Italian extensions live in the `it-sdi` addon (`addons/it/sdi/extensions.go`) using the `it-sdi-*` convention (and that file already notes the keys are "defined as part of the FatturaPA specification"). There's no `it-fatturapa` addon in the repo today.

My preference would be to add them to the existing `it-sdi` addon as `it-sdi-liquidation-state` (values `LS`/`LN`) and `it-sdi-shareholder-state` (values `SU`/`SM`). It's consistent with the current naming, it's less work than standing up a new addon, and it should land faster. Happy to go the new-addon route instead if you and @rilla would rather keep `IscrizioneREA` fields separate.

Once the naming is settled I'll open a small PR with the two extension definitions (enumerated values + descriptions), then rework invopop/gobl.fatturapa#68 to read them from `registration.ext` instead of the removed fields.

pmenendz:
Hi @aroundthfur ,

It is indeed it-sdi. It should follow the addon convention

aroundthfur:
Thanks @samlown and @pmenendz. I've opened #864, which implements the agreed approach: the two values are now `it-sdi-liquidation-state` and `it-sdi-shareholder-state` extensions on the `it-sdi` addon, used via the `org.Registration.ext` field from #863.

Closing this one in favour of #864.

PR #857, opened 2026-06-11
Keeping the MVA suffix on Norwegian VAT identities

CarlosMari:
Norwegian VAT numbers take the form `NO<orgnr>MVA` — the MVA suffix is part of
the identifier, and the Peppol national rule NO-R-001 rejects supplier VAT
numbers without it ("For Norwegian suppliers, a VAT number MUST be the country
code prefix NO followed by a valid Norwegian organization number (nine numbers)
followed by the letters MVA"). The NO regime's normalizer was stripping the
suffix, so every serialized Norwegian VAT identity went out as `NO<orgnr>` and
failed validation at receiving access points (reported via support).

## Changes

- `normalizeTaxIdentity` now keeps the suffix and appends it to identities
  given as a bare organisation number, making `<orgnr>MVA` the canonical stored
  code. Serializers compose the country prefix as usual, so the wire form is
  `NO<orgnr>MVA` everywhere with no downstream changes.
- Validation requires the suffixed form (mod-11 checksum on the nine digits).
- Organisation-number identities (`org.Identity`) are unchanged — the orgnr
  itself carries no suffix.
- NO examples regenerated.

## Tests

Normalization table covers bare, suffixed, lowercase, spaced, dashed and
NO-prefixed inputs; validation rejects bad checksums, wrong lengths,
double-suffixed and non-numeric codes. Verified end-to-end against the Peppol
2025.5 schematron (phive): invoices and credit notes with the suffixed form
pass; the previous stripped form fails NO-R-001 fatally. Full suite and
`mage check` green.

## Pre-Review Checklist

### DIFF
--- a/regimes/no/tax_identity.go
+++ b/regimes/no/tax_identity.go
@@ -18,8 +18,8 @@ func taxIdentityRules() *rules.Set {
 	return rules.For(new(tax.Identity),
 		rules.When(tax.IdentityIn(CountryCode),
 			rules.Field("code",
-				rules.AssertIfPresent("01", "invalid organisasjonsnummer",
-					is.Func("valid mod-11 org number", isValidOrgNumber),
+				rules.AssertIfPresent("01", "invalid Norwegian VAT number",
+					is.Func("valid MVA-suffixed mod-11 org number", isValidVATCode),
 				),
 			),
 		),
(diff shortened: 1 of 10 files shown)


regimes/no/tax_identity.go
```diff
@@ -18,20 +18,34 @@ func taxIdentityRules() *rules.Set {
...
+// ensures the "MVA" suffix that Norwegian VAT numbers carry (e.g.
+// "NO 923 456 783 MVA"): a VAT identity given as a bare organisation number
+// gains the suffix, so the serialized form is always NO<orgnr>MVA as the
+// EHF and Peppol national rules require.
 func normalizeTaxIdentity(tID *tax.Identity) {
 	tax.NormalizeIdentity(tID)
-	tID.Code = cbc.Code(strings.TrimSuffix(string(tID.Code), "MVA"))
+	if tID.Code != "" && !strings.HasSuffix(string(tID.Code), "MVA") {
+		tID.Code += "MVA"
+	}
```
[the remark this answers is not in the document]
CarlosMari: Org number is tested without MVA code

samlown:
I think this one needs a bit more discussion. As a rule in GOBL we try to avoid repeated codes and instead focus on the simplest normalized form. The "MVA" doesn't add any meaning to the code and could be considered just a visual component added during conversion to PDF or the Peppol Participant ID, the same as with the country code.

regimes/no/tax_identity.go
```diff
@@ -18,20 +18,34 @@ func taxIdentityRules() *rules.Set {
...
+}
+
+// isValidVATCode reports whether the value is a valid Norwegian VAT number
+// code: a mod-11 organisation number followed by the "MVA" suffix.
+func isValidVATCode(value any) bool {
+	code, ok := value.(cbc.Code)
+	if !ok || !strings.HasSuffix(string(code), "MVA") {
+		return false
+	}
+	return isValidOrgNumber(cbc.Code(strings.TrimSuffix(string(code), "MVA")))
```
samlown: There is a pattern here. It might be useful to add `HasSuffix` and `TrimSuffix` (plus `HasPrefix` and `TrimePrefix`) to the cbc.Code itself so that we don't have to repeat so much boilerplate.

samlown (approved):
Approved, but for bonus points, I think it would be useful to add some helper methods to `cbc.Code`.

regimes/no/tax_identity.go
CarlosMari: Fixed, added the different methods

PR #849, opened 2026-06-04
Feat/regime no

CarlosMari:
Closes: APP-567

## Summary

Adds the Norwegian (NO) tax regime. Supersedes #728 by @0xjgv — this branch
takes that work, rebases it onto current `main`, migrates it to the current
`rules` validation framework, and trims it to what Norwegian law actually
requires.

Co-authored with @0xjgv (original commits preserved in the history).

## What's included

- **VAT (Merverdiavgift / MVA)** with the four rates in force per the
  *Stortingsvedtak om merverdiavgift for 2026*:
  - 25% general
  - 15% reduced (food, water, wastewater — mval § 5-2)
  - 12% super-reduced (transport, accommodation, cinema, broadcasting)
  - 11.11% special (wild marine resources — mval § 5-8)
  - zero / exempt / reverse-charge / export / outside-scope via
    `tax.GlobalVATKeys()`, as other regimes do.
- **Organisasjonsnummer** (`org.Identity` type `ON`) and tax-identity
  validation: normalization strips the `NO` prefix and `MVA` suffix, then
  validates the 9-digit mod-11 check digit (per Brønnøysundregistrene).
- **Invoice validation** (only the Norway-specific rules; GOBL core already
  covers the universal fields):
  - supplier must carry a tax ID **or** an `ON` org identity
    (bokføringsforskriften § 5-1-2) — a VAT registration is not required, so
    sub-NOK-50,000 businesses can still invoice;
  - customer required on non-simplified invoices (§ 5-1-2);
  - credit notes must reference the preceding document (§ 5-2-7).
- **Simplified invoices** relax the customer requirement.
- **Corrections**: credit note (kreditnota) only — Norwegian law has no
  debit-note concept (§ 5-2-7).
- **Reverse-charge** scenario note (Omvendt avgiftsplikt).
- Examples + tests, generated regime/rules data, CHANGELOG.

## Deliberately deferred to a future EHF / SAF-T addon

These are real legal requirements but don't belong in the regime layer
(they depend on the supplier's legal form, which GOBL core doesn't model, or
on currency conversion, which gobl now handles in addons):

- the word **"Foretaksregisteret"** and the seller's head-office **address**,
  required only for AS/ASA and foreign branches (§ 5-1-2 / foretaksregisterloven
  § 10-2);
- stating the **VAT amount in NOK** on foreign-currency invoices (§ 5-1-1 nr. 6).

## How to test

```bash
go test -race ./...
[...]

PR #858, opened 2026-06-12
Add Luxembourg (LU) tax regime

Lluc24:
Adds Luxembourg (`LU`) as a new tax regime.

## What's included

- **`regimes/lu`** — regime definition following the structure of the recently merged Norway regime (#849):
  - **Four VAT rates** with full `Since`-dated history: standard 17%, intermediate ("parking") 14%, reduced 8%, super-reduced 3%. The history includes the temporary one-point reduction in force throughout 2023 (Law of 26 October 2022, Mémorial A No. 534) — the standard, intermediate, and reduced rates dropped to 16/13/7% and were restored on 2024-01-01; the super-reduced rate was excluded from the measure. An invoice dated within 2023 automatically calculates with the temporary rates.
  - **TVA number validation**: 8 digits where the last two must equal the first six mod 89, plus normalization of the `LU` prefix, spaces, and dashes.
  - **RCS identity** (`RCS`): Luxembourg Business Registers company number, format `<section letter> + 1–6 digits` (e.g. `B263475`). Any single uppercase section letter is accepted — the published list of sections (A sole traders, B commercial companies, C GIEs, E civil companies, …) has grown over time, so a closed list risks rejecting valid registrations.
  - **Invoice rule**: the supplier must carry either a TVA code or an RCS identity. Note this is a structural baseline so businesses below the VAT registration threshold can still issue valid invoices — not a claim that an RCS number legally substitutes for the TVA number, which VAT-registered suppliers must always state.
  - **Corrections**: credit notes and debit notes, both recognised in Luxembourg practice.
- **`examples/lu`** — four documents with golden outputs: standard B2B invoice (17%), multi-rate invoice (17/8/3%), a 2023-dated invoice demonstrating the historical 16%/7% rates, and a credit note.
- Generated data (`data/regimes/lu.json`, `data/rules/lu.json`, regime-code schema) and a CHANGELOG entry.

## Scope decisions

- **No addon.** Luxembourg has no domestic B2B e-invoicing mandate; B2G uses Peppol/EN 16931, already covered by the existing `eu-en16931-v2017` addon. A `lu-*` addon would be speculative at this point (pending ViDA).
- **No withholding taxes, OSS/IOSS, or VAT-group handling** — respectively income-tax mechanisms, EU-level machinery, and registration arrangements invisible at the single-invoice level.
- **Pre-2015 rate history** (15/12/6/3%) is anchored at a conservative 1992-01-01 baseline; the values are well documented but I could not verify the exact original effective dates under the amended 1979 law against primary sources. Happy to adjust if you have a preferred convention for historical anchors.

## Sources

- AED – VAT rates: https://www.aed.public.lu/en/tva/taux-tva.html
- Law of 26 October 2022 (temporary reduction): https://legilux.public.lu/eli/etat/leg/loi/2022/10/26/a534/jo
- Law of 19 December 2014 (2015 increase): https://www.legilux.public.lu/eli/etat/leg/loi/2014/12/19/n1/jo
- European Commission – VAT rates: https://taxation-customs.ec.europa.eu/taxation/vat/eu-vat-rules/vat-rates_en
- Luxembourg Business Registers (RCS): https://www.lbr.lu

Structured sources are also embedded in the regime and category definitions.

## Pre-Review Checklist

- [x] Added thorough tests with at **least** 90% code coverage. *(97.3% of statements)*

### DIFF
--- a/data/rules/lu.json
+++ b/data/rules/lu.json
@@ -66,8 +66,8 @@
               "assert": [
                 {
                   "id": "GOBL-LU-BILL-INVOICE-01",
-                  "desc": "invoice LU supplier must have either a TVA tax ID code or an identity of type 'RCS'",
-                  "tests": "has TVA code or RCS identity"
+                  "desc": "invoice LU supplier must have a TVA tax ID code or an RCS identity",
+                  "tests": "has TVA tax ID code, or has RCS identity"
                 }
               ]
             }
--- a/data/rules/lu.json
+++ b/data/rules/lu.json
@@ -1,0 +1,79 @@
+{
+  "id": "GOBL-LU",
+  "package": "lu",
+  "subsets": [
+    {
+      "id": "GOBL-LU-TAX-IDENTITY",
+      "object": "tax.Identity",
+      "subsets": [
+        {
+          "guard": "code in [LU]",
+          "subsets": [
+            {
+              "field": "code",
+              "subsets": [
+                {
+                  "guard": "present",
+                  "assert": [
+                    {
+                      "id": "GOBL-LU-TAX-IDENTITY-01",
+                      "desc": "invalid Luxembourg TVA number",
+                      "tests": "valid mod-89 TVA code"
+                    }
+                  ]
+                }
+              ]
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "id": "GOBL-LU-ORG-IDENTITY",
+      "object": "org.Identity",
+      "subsets": [
+        {
+          "guard": "context: regime in [LU]",
+          "subsets": [
+            {
+              "guard": "type in [RCS]",
+              "subsets": [
+                {
+                  "field": "code",
+                  "assert": [
+                    {
+                      "id": "GOBL-LU-ORG-IDENTITY-01",
+                      "desc": "invalid RCS number",
+                      "tests": "valid RCS format"
+                    }
+                  ]
+                }
+              ]
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "id": "GOBL-LU-BILL-INVOICE",
+      "object": "bill.Invoice",
+      "subsets": [
+        {
+          "guard": "context: regime in [LU]",
+          "subsets": [
+            {
+              "field": "supplier",
+              "assert": [
+                {
+                  "id": "GOBL-LU-BILL-INVOICE-01",
+                  "desc": "invoice LU supplier must have either a TVA tax ID code or an identity of type 'RCS'",
+                  "tests": "has TVA code or RCS identity"
+                }
+              ]
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
--- a/data/rules/lu.json
+++ b/data/rules/lu.json
@@ -17,8 +17,28 @@
                   "assert": [
                     {
                       "id": "GOBL-LU-TAX-IDENTITY-01",
-                      "desc": "invalid Luxembourg TVA number",
-                      "tests": "valid mod-89 TVA code"
+                      "desc": "Luxembourg TVA number must be exactly 8 digits",
+                      "tests": "length between 8 and 8"
+                    }
+                  ]
+                },
+                {
+                  "guard": "present",
+                  "assert": [
+                    {
+                      "id": "GOBL-LU-TAX-IDENTITY-02",
+                      "desc": "Luxembourg TVA number must contain only digits",
+                      "tests": "digit"
+                    }
+                  ]
+                },
+                {
+                  "guard": "present",
+                  "assert": [
+                    {
+                      "id": "GOBL-LU-TAX-IDENTITY-03",
+                      "desc": "invalid Luxembourg TVA number: mod-89 checksum mismatch",
+                      "tests": "valid mod-89 checksum"
                     }
                   ]
                 }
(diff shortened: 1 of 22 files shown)


regimes/lu/bill_invoices.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/invopop/gobl/tax"
+)
+
+// billInvoiceRules requires the supplier to carry at least one official
+// identifier. Note that the RCS number is a company-registry reference, not
+// a tax identifier: a VAT-registered supplier is still legally required to
+// state its TVA number on invoices. Accepting an RCS identity as an
+// alternative is a deliberate baseline so that businesses below the VAT
+// registration threshold (e.g. small franchised businesses) can still issue
+// valid GOBL invoices; it is not a claim of legal equivalence.
```
samlown: A bit of Opus 4.X model slop here :-) can probably be reduced.

regimes/lu/bill_invoices.go
```diff
@@ -0,0 +1,49 @@
...
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(l10n.LU.Tax())),
+			rules.Field("supplier",
+				rules.Assert("01",
+					fmt.Sprintf("invoice LU supplier must have either a TVA tax ID code or an identity of type '%s'", IdentityTypeRCS),
+					is.Func("has TVA code or RCS identity", hasVATCodeOrRCSIdentity),
```
samlown: This should be simplified to use the `is.AnyOf` method instead of the function test for readability.

samlown (changes requested):
Great PR! A few style changes I think would help.

regimes/lu/bill_invoices.go
Lluc24: Good catch, trimmed it down to two lines

regimes/lu/bill_invoices.go
Lluc24: Done, switched to is.AnyOf with two named is.Func checks

regimes/lu/bill_invoices.go
```diff
@@ -0,0 +1,42 @@
...
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(l10n.LU.Tax())),
+			rules.Field("supplier",
+				rules.Assert("01",
+					"invoice LU supplier must have a TVA tax ID code or an RCS identity",
+					is.AnyOf(
+						is.Func("has TVA tax ID code", hasTaxIDCode),
+						is.Func("has RCS identity", hasRCSIdentity),
```
samlown: There are helpers to do this, e.g.: https://github.com/invopop/gobl/blob/main/org/identity.go#L100

regimes/lu/org_identities.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	id.Code = cbc.NormalizeAlphanumericalCode(id.Code)
+}
+
+func isValidRCSCode(value any) bool {
+	code, ok := value.(cbc.Code)
+	if !ok || code == "" {
+		return false
+	}
+	return rcsRegexp.MatchString(code.String())
```
samlown: If we're only doing a regexp, you can use `is.Matches` for this.

regimes/lu/tax_identity.go
```diff
@@ -0,0 +1,59 @@
...
+// Source: https://www.aed.public.lu/en/tva/numero-tva.html
+func isValidTVACode(value any) bool {
+	code, ok := value.(cbc.Code)
+	if !ok {
+		return false
+	}
+	return validateTVACode(code) == nil
+}
+
+func validateTVACode(code cbc.Code) error {
```
samlown: This method can be simplified to just a true or false response specifically for the checksum. The length and numerical check can be performed directly in the rules. This makes it easier to debug as the error messages will never be provided to users.

samlown (changes requested):
Still a few issues with the approaches for handling test functions: we should be re-using as much as possible.

regimes/lu/bill_invoices.go
Lluc24: Will update

regimes/lu/org_identities.go
Lluc24: Agreed, I'll swap `is.MatchesRegexp(rcsRegexp)` for `is.Matches(rcsRegexp)`.

regimes/lu/tax_identity.go
Lluc24: I'll strip the defensive returns and make it a pure checksum predicate

data/rules/lu.json
```diff
@@ -0,0 +1,79 @@
...
+              "subsets": [
+                {
+                  "guard": "present",
+                  "assert": [
+                    {
+                      "id": "GOBL-LU-TAX-IDENTITY-01",
+                      "desc": "invalid Luxembourg TVA number",
+                      "tests": "valid mod-89 TVA code"
+                    }
+                  ]
```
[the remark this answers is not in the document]
Lluc24: You're right, the JSON schemas were stale and are now regenerated with `go generate`

data/rules/lu.json
```diff
@@ -0,0 +1,79 @@
...
+          "subsets": [
+            {
+              "field": "supplier",
+              "assert": [
+                {
+                  "id": "GOBL-LU-BILL-INVOICE-01",
+                  "desc": "invoice LU supplier must have a TVA tax ID code or an RCS identity",
+                  "tests": "has TVA tax ID code, or has RCS identity"
+                }
+              ]
```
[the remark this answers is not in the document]
Lluc24: Same, solved stale JSON schemas with `go generate`

PR #860, opened 2026-06-12
cef: adopt VATEX code list version 8.0

alvarolivie:
- `data/catalogues/cef.json`: the `cef-vatex` extension enumerates the 88 codes of the official VATEX code list version 8.0, replacing the shape-only pattern. Version 8 added the French CTC codes that motivated the pattern (`VATEX-FR-CGI*`, `VATEX-FR-AE`, `VATEX-FR-298SEXDECIESA`) plus five new EU codes, so enumerating is possible again and stricter.
- Each code carries the list's `nationality`, `deprecated` and `first-version` columns as metadata, `remark` and `last-version` where present, and its context of exemption as the description. The extension records the list `version` and its sources.
- The 16 `VATEX-SA-*` codes are removed — never part of the CEF list; `gobl.sa.zatca` will define them in its own catalogue.
- Tests: code membership, the version, and per-code metadata in `catalogues/cef`.

## Pre-Review Checklist

- [ ] Read the CONTRIBUTING.md guide.
- [ ] Modified or created example GOBL documents to show my changes in use, if appropriate.


samlown (changes requested):
As discussed, the best option here is to ensure the cef.json only contains codes that are prepared by the CEF (Connecting Europe Facility). Codes from other regions, such as Zatca (Saudi Arabia), should be handled by their own catalogues in addons.

samlown (approved):
LGTM!

PR #867, opened 2026-06-17
pkg/examples: normalize path separators before skip-list matching

### LINKED ISSUE #868: pkg/examples: skip-list matching silently fails on Windows
Sources in `pkg/examples/examples.go` walks the filesystem with `filepath.Walk` and checks each path against `defaultSkip` fragments (`/out/`, `.out.`, `/.git`) using a plain `strings.Contains`. On Windows, `filepath.Walk` returns backslash-separated paths (`examples\nz\out\invoice.json`), so the forward-slash fragments never match and generated output files in `out/` are included in the discovered sources instead of being skipped.

This causes `TestConvertExamplesToJSON` to fail on Windows: it tries to convert already-generated JSON snapshots as if they were source examples.

Fix: normalize the path with `filepath.ToSlash` before the `strings.Contains` check, so skip-list matching is platform-independent.

Fixed by PR #867

pedrorfdez:
- pkg/examples: use `filepath.ToSlash` before skip-list matching so Sources correctly excludes `/out/` and `/.git` paths on Windows, where `filepath.Walk` returns backslash-separated paths.
- Fixes #868 

## Pre-Review Checklist

- [N/A] Added links to the source of the changes in tax regimes or addons, either structured or in the comments.


samlown (approved):
Nice! Thanks for this!

PR #871, opened 2026-06-21
Add Estonia (EE) tax regime

sandragc8:
## Summary

This PR adds a new tax regime for **Estonia (EE)**. The motivation is the scenario in the assignment: a company already invoicing in one country opens a subsidiary in Estonia, which GOBL did not yet cover.

I used the existing EU regimes (Austria and Ireland) as structural references, since Estonia is an EU member state with a VAT system and the euro.

## What's included

- **VAT category** with four rates:
  - Standard `general` rate — **24%** (current), with historical values kept so past-dated invoices still resolve: **22%** since 2024-01-01 and **20%** since 2009-07-01.
  - `reduced` — **13%** (accommodation services).
  - `super-reduced` — **9%** (books, press publications, certain medicines).
  - `zero` — **0%** (exports, intra-community supply).
- **KMKR VAT identity validation** (`tax_identity.go`): `EE` + 9 digits, validated after the country prefix is stripped during normalization. Includes the mod-10 weighted checksum (weights `[3,7,1,3,7,1,3,7]`).
- **Credit notes** enabled as invoice corrections.
- Regime registered in `regimes/regimes.go`, generated `data/regimes/ee.json`, and a `CHANGELOG.md` entry.

## Tests

- `regimes/ee/tax_identity_test.go` — table tests with real valid KMKR numbers and invalid cases (format, length, un-normalized prefix, bad checksum).
- `regimes/ee/ee_test.go` — sanity check of the regime definition.
- The global `regimes_test.go` validates the new `RegimeDef` automatically.

## Sources

VAT rates and effective dates come from the Estonian Tax and Customs Board (EMTA): https://www.emta.ee/en/business-client/taxes-and-payment/value-added-tax/vat-rates-and-supply-exempt-tax/standard-vat-rate

## Scope — what I deliberately left out

To keep the change focused and reviewable, I limited it to what the law requires for correct invoicing, in line with the "minimum complexity for legal correctness" principle:

- **No e-invoicing addon** (PEPPOL / national CTC). Estonia supports e-invoicing, but that belongs in a separate addon rather than the base regime.
- **No transitional 20% provisions** (cash-accounting / long-term property contracts valid until mid-2025). These depend on time-of-supply rules, not on a standing rate, so modelling them as an active rate would be misleading.
- **No full exemptions catalogue** (financial, insurance, real estate). The global VAT keys (`exempt`, `reverse-charge`, etc.) already cover the common cases.
- **No live VIES/EMTA lookup** for registration status — that's an application-layer concern; the regime validates format and checksum offline, consistent with the other regimes (AT, IE).

Happy to extend any of the above if you'd prefer it in scope.

### DIFF
--- a/regimes/ee/ee.go
+++ b/regimes/ee/ee.go
@@ -1,0 +1,71 @@
+// Package ee provides the Estonian tax regime.
+package ee
+
+import (
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/currency"
+	"github.com/invopop/gobl/i18n"
+	"github.com/invopop/gobl/norm"
+	"github.com/invopop/gobl/pkg/here"
+	"github.com/invopop/gobl/rules"
+	"github.com/invopop/gobl/tax"
+)
+
+// CountryCode is the tax country code for Estonia.
+const CountryCode = "EE"
+
+func init() {
+	tax.RegisterRegimeDef(New())
+	rules.Register("ee", rules.GOBL.Add(CountryCode), taxIdentityRules())
+	norm.Register(
+		norm.When(tax.IdentityIn(CountryCode), norm.For(func(id *tax.Identity) { tax.NormalizeIdentity(id) })),
+	)
+}
+
+// New provides the tax regime definition for Estonia.
+func New() *tax.RegimeDef {
+	return &tax.RegimeDef{
+		Country:   CountryCode,
+		Currency:  currency.EUR,
+		TaxScheme: tax.CategoryVAT,
+		Name: i18n.String{
+			i18n.EN: "Estonia",
+			i18n.ET: "Eesti",
+		},
+		Description: i18n.String{
+			i18n.EN: here.Doc(`
+				Estonia's tax system is administered by the Estonian Tax and Customs
+				Board (Maksu- ja Tolliamet, EMTA). As an EU member state, Estonia follows
+				the EU VAT Directive with standard and reduced rates.
+
+				VAT (Käibemaks) is applied to most goods and services. The standard rate
+				is 24% since 1 July 2025, raised from 22% (in force since 1 January 2024),
+				which in turn replaced the long-standing 20% rate in force since 2009.
+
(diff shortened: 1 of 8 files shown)


regimes/ee/ee.go
```diff
@@ -0,0 +1,71 @@
...
+		},
+		Description: i18n.String{
+			i18n.EN: here.Doc(`
+				Estonia's tax system is administered by the Estonian Tax and Customs
+				Board (Maksu- ja Tolliamet, EMTA). As an EU member state, Estonia follows
+				the EU VAT Directive with standard and reduced rates.
+
+				VAT (Käibemaks) is applied to most goods and services. The standard rate
+				is 24% since 1 July 2025, raised from 22% (in force since 1 January 2024),
+				which in turn replaced the long-standing 20% rate in force since 2009.
```
samlown, line 44: We usually try to avoid adding rates in the description as they're already defined in the tax categories and may become desynchronised.

regimes/ee/tax_categories.go
```diff
@@ -0,0 +1,98 @@
...
+					},
+					{
+						Since:   cal.NewDate(2009, 7, 1),
+						Percent: num.MakePercentage(200, 3),
+					},
+				},
+			},
+			{
+				// Higher reduced rate, currently applied to accommodation services.
+				Rate: tax.RateReduced,
```
samlown, line 60: This I think feels more like an "intermediate" rate as opposed to reduced.

regimes/ee/tax_categories.go
```diff
@@ -0,0 +1,98 @@
...
+				},
+				Values: []*tax.RateValueDef{
+					{
+						Percent: num.MakePercentage(130, 3),
+					},
+				},
+			},
+			{
+				// Lower reduced rate for books, press publications and certain medicines.
+				Rate: tax.RateSuperReduced,
```
samlown, line 73: This is probably "reduced", unless you see evidence for something different.

samlown (changes requested):
Thanks for this! I'd just check the copy in the description, and the rate assignment.

samlown:
We also lost the PR checklist here.

PR #875, opened 2026-06-25
Decouple totals calculator test from addon packages

rilla:
## Context

The tax-regime addons are moving out of core into standalone modules (`it-sdi` → `gobl.it.sdi`, and `es-tbai` / `pt-saft` to follow). `tax/totals_calculator_test.go` imported three of them (`addons/es/tbai`, `addons/it/sdi`, `addons/pt/saft`) — but only to borrow their extension-key constants as fixture data.

The totals calculator is addon-agnostic: the test calls `tax.TotalCalculator.Calculate` directly (no invoice, no addon, no validation), and the calculator groups rates by whatever extension is present, regardless of which regime or addon owns the key. So the dependency was incidental.

## Changes

- Replace the borrowed addon constants (`tbai.ExtKeyExempt`, `sdi.ExtKeyRetained`, `saft.ExtKeyTaxRate`) with local placeholder keys, and drop the three addon imports.
- Assertions are unchanged — the calculator only cares that extensions differ, not what they are.

## Next

Removes a blocker for externalising those addons: their extraction no longer has to touch this core test.

### DIFF
--- a/tax/totals_calculator_test.go
+++ b/tax/totals_calculator_test.go
@@ -4,9 +4,6 @@ import (
 	"encoding/json"
 	"testing"
 
-	"github.com/invopop/gobl/addons/es/tbai"
-	"github.com/invopop/gobl/addons/it/sdi"
-	"github.com/invopop/gobl/addons/pt/saft"
 	"github.com/invopop/gobl/cal"
 	"github.com/invopop/gobl/cbc"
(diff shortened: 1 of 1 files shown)


rilla:
Folding this into the upcoming it-sdi core-externalisation PR instead, so the addon removal and this test decoupling land together on one appropriately-named branch.

PR #893, opened 2026-07-03
Add Slovenia (SI) tax regime

daesparza02:
This PR adds Slovenia (SI) as a new tax regime.

I picked Slovenia because it wasn't supported yet and fits GOBL's model well: an EU member
state using EUR, with tax identifiers that carry real modulo-11 check digits, so the regime
can do genuine validation rather than only format checks. I used the Netherlands (`nl`)
regime as a structural reference, since it's close to Slovenia: EU member, EUR, and a VAT
identity with a checksum.

## What's included

- **Base regime:** EUR, `Europe/Ljubljana`, the standard invoice scenarios, and credit notes for corrections.
- **VAT rates:** the standard rate with its history (20% since 2002, 22% from 1 July 2013), plus the current reduced rate (9.5%) and the special reduced rate of 5% for books, newspapers and periodicals.
- **Tax ID validation:** the modulo-11 check-digit algorithm, applied to two identifiers — the VAT number / davčna številka (8 digits) and the domestic registration number / matična številka (10 digits).
- **Invoice rule:** the supplier must carry a tax ID code. The registration number identifies a company but is not a substitute for the VAT number on an invoice, so the rule checks specifically for the tax ID (ZDDV-1, Article 82).

## Scope and design notes

A few things I decided deliberately, with the reasoning in case it helps the review:

- I modelled the full date history only for the standard rate, to exercise the `Since` mechanism. The reduced rates have been stable for years, so I kept only their current values rather than pad the base regime with historical data — happy to add the full history if you'd prefer.
- The modulo-11 logic lives in one place (`tax_identity.go`) and takes the weights as a parameter, since the two identifiers share the same scheme with different weights. The registration-number check strips its three-digit branch suffix (which has no check digit of its own) and reuses it.
- One Slovenian detail worth flagging: a remainder of 0 is never assigned (no valid number produces it), so the shared validator rejects it explicitly.
- I kept the scope to tax and identity rules — Slovenia's own e-invoicing formats and networks are a separate kind of extension in GOBL, so I left those out.

## Testing

- Table-driven tests across the package: 5 test functions, 29 sub-cases, covering valid and invalid inputs.
- The official sources publish the number formats but not the check-digit formulas, so I verified each checksum against a real, public identifier from the same company (Krka d.d.: VAT number `SI82646716` and registration number `5043611000`) and exercised every branch of both algorithms, including the `remainder == 0` case that can never yield a valid number.
- `gofmt`, `go vet` and `golangci-lint` are clean.

## Sources

- VAT rates — Financial Administration of the Republic of Slovenia (FURS), and the European Commission rates table
  https://www.fu.gov.si/en/taxes_and_other_duties/areas_of_work/value_added_tax_vat
  https://taxation-customs.ec.europa.eu/system/files/2021-06/vat_rates_en.pdf
- Invoice requirements — SPOT business portal (ZDDV-1, Article 82)
  https://spot.gov.si/en/info/accountancy/invoicing
- Tax number format — FURS
  https://www.fu.gov.si/en/taxes_and_other_duties/work_with_us/entry_into_the_tax_register_and_tax_number
- Registration number (matična številka) format — PISRS, Article 11 of the Business Register decree
  https://pisrs.si/Pis.web/pregledPredpisa?id=URED7599
- Checksum algorithms — the official registers document the formats but not the formulas, so I
  based them on a widely used open-source implementation (python-stdnum) and verified them
  against the public identifiers above:
  https://github.com/arthurdejong/python-stdnum

## Pre-Review Checklist

### DIFF
--- a/regimes/si/bill_invoices.go
+++ b/regimes/si/bill_invoices.go
@@ -2,7 +2,6 @@ package si
 
 import (
 	"github.com/invopop/gobl/bill"
-	"github.com/invopop/gobl/org"
 	"github.com/invopop/gobl/rules"
 	"github.com/invopop/gobl/rules/is"
 	"github.com/invopop/gobl/tax"
@@ -17,15 +16,13 @@ func billInvoiceRules() *rules.Set {
 		rules.When(
 			is.InContext(tax.RegimeIn(CountryCode)),
 			rules.Field("supplier",
-				rules.Assert("01", "invoice SI supplier must have a tax ID code",
-					is.Func("has tax ID code", hasSupplierTaxIDCode),
+				rules.Field("tax_id",
+					rules.Assert("01", "invoice SI supplier must have a tax ID", is.Present),
+					rules.Field("code",
+						rules.Assert("02", "invoice SI supplier tax ID must have a code", is.Present),
+					),
 				),
 			),
 		),
 	)
-}
-
-func hasSupplierTaxIDCode(value any) bool {
-	party, _ := value.(*org.Party)
-	return party != nil && party.TaxID != nil && party.TaxID.Code != ""
 }
--- a/regimes/si/bill_invoices.go
+++ b/regimes/si/bill_invoices.go
@@ -1,0 +1,31 @@
+package si
+
+import (
+	"github.com/invopop/gobl/bill"
+	"github.com/invopop/gobl/org"
+	"github.com/invopop/gobl/rules"
+	"github.com/invopop/gobl/rules/is"
+	"github.com/invopop/gobl/tax"
+)
+
+// Article 82 of the Slovenian VAT Act (ZDDV-1) requires invoices to state
+// the VAT identification number under which the taxable person supplied
+// the goods or services.
+// Source: https://spot.gov.si/en/info/accountancy/invoicing
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(CountryCode)),
+			rules.Field("supplier",
+				rules.Assert("01", "invoice SI supplier must have a tax ID code",
+					is.Func("has tax ID code", hasSupplierTaxIDCode),
+				),
+			),
+		),
+	)
+}
+
+func hasSupplierTaxIDCode(value any) bool {
+	party, _ := value.(*org.Party)
+	return party != nil && party.TaxID != nil && party.TaxID.Code != ""
+}
--- a/regimes/si/org_identities.go
+++ b/regimes/si/org_identities.go
@@ -13,13 +13,13 @@ const (
 	// IdentityTypeMaticna represents the Slovenian registration number
 	// (matična številka) assigned by AJPES to every entity entered in
 	// the Slovenian Business Register (e.g. "5043611000").
-	IdentityTypeMaticna cbc.Code = "MATICNA"
+	IdentityTypeMaticna cbc.Code = "MŠ"
 )
 
-// registrationNumberLen is the length of a registration number: six
-// digits, a modulo-11 check digit, and a three-digit suffix identifying
-// the business unit ("000" for the head office).
-const registrationNumberLen = 10
+// registrationNumberPattern matches a registration number (matična
+// številka): six base digits, a modulo-11 check digit, and a three-digit
+// suffix identifying the business unit ("000" for the head office).
+const registrationNumberPattern = `^\d{10}$`
 
 // registrationNumberMultipliers are the weights applied to the first six
 // digits of the registration number to calculate the check digit.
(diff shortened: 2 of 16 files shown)


regimes/si/bill_invoices.go
```diff
@@ -0,0 +1,31 @@
...
+// the VAT identification number under which the taxable person supplied
+// the goods or services.
+// Source: https://spot.gov.si/en/info/accountancy/invoicing
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(CountryCode)),
+			rules.Field("supplier",
+				rules.Assert("01", "invoice SI supplier must have a tax ID code",
+					is.Func("has tax ID code", hasSupplierTaxIDCode),
```
samlown: Its better to use direct rule checks here instead of a helper.

regimes/si/org_identities.go
```diff
@@ -0,0 +1,79 @@
...
+var identityDefinitions = []*cbc.Definition{
+	{
+		Code: IdentityTypeMaticna,
+		Name: i18n.String{
+			i18n.EN: "Registration Number",
+			i18n.SL: "Matična številka",
+		},
+		Sources: []*cbc.Source{
+			{
+				Title: i18n.NewString("Uredba o vodenju in vzdrževanju Poslovnega registra Slovenije"),
```
samlown: Missing English translation.

regimes/si/org_identities.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/invopop/gobl/rules"
+	"github.com/invopop/gobl/rules/is"
+	"github.com/invopop/gobl/tax"
+)
+
+const (
+	// IdentityTypeMaticna represents the Slovenian registration number
+	// (matična številka) assigned by AJPES to every entity entered in
+	// the Slovenian Business Register (e.g. "5043611000").
+	IdentityTypeMaticna cbc.Code = "MATICNA"
```
samlown: Best to avoid local names in variables, there is likely to be a good translation. I'd also question the `MATICNA` code itself, it looks like a name that's been capitalized as opposed to a name, I'd expect there to be some type of abbreviated version.

regimes/si/org_identities.go
```diff
@@ -0,0 +1,79 @@
...
+
+func orgIdentityRules() *rules.Set {
+	return rules.For(new(org.Identity),
+		rules.When(
+			is.InContext(tax.RegimeIn(CountryCode)),
+			rules.When(
+				org.IdentityTypeIn(IdentityTypeMaticna),
+				rules.Field("code",
+					rules.Assert("01", "identity code for type MATICNA must be valid",
+						is.Func("valid", isValidRegistrationNumber),
```
samlown: Checks here could be split up to help users differentiate between length and check digit.

regimes/si/tax_identity.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil
+	}
+	if len(code) != taxCodeLen {
+		return errors.New("invalid length")
+	}
+	if code[0] == '0' {
+		return errors.New("invalid format")
+	}
+	return validateMod11(code, taxCodeMultipliers)
+}
```
samlown: These errors will never reach anywhere, so the checks should probably be split into separate rules.

regimes/si/tax_identity.go
```diff
@@ -0,0 +1,94 @@
...
+// validateMod11 checks that code is made up of digits and that its last
+// digit is the modulo-11 check digit of the preceding ones, weighted with
+// the given multipliers. It expects exactly len(weights)+1 characters.
+//
+// Both Slovenian identifiers (the tax number and the registration number)
+// share this scheme with different weights: a remainder of 0 is never
+// assigned (no valid number produces it), a remainder of 1 means a check
+// digit of 0, and any other remainder means a check digit of 11 minus the
+// remainder.
+func validateMod11(code cbc.Code, weights []int) error {
```
samlown: Again, for a private method, the errors returned are not useful.

samlown (changes requested):
Thanks! A few style changes for the rule handling required.

regimes/si/bill_invoices.go
daesparza02: Done — replaced the helper with direct nested field rules using is.Present on tax_id and code.

regimes/si/org_identities.go
daesparza02: Fixed, the source title now carries both English and Slovenian.

regimes/si/org_identities.go
daesparza02: I changed the code from the spelled-out word to MŠ, the official Slovenian abbreviation for matična številka. f you'd rather avoid the non-ASCII character, I'm happy to use an ASCII form (MS).

regimes/si/org_identities.go
daesparza02: Done, split into three separate rules: presence, format and the check digit. Each failure now reports its own code and message.

regimes/si/tax_identity.go
daesparza02: Done, removed validateTaxCode and split it into two rules: a format check (01) using a regex that covers length, digits and the non-zero first digit, and the check digit (02).

regimes/si/tax_identity.go
daesparza02: Done, validMod11 now returns a bool rather than building error values that were discarded; the user-facing message lives in the check-digit rule.

daesparza02:
Thanks for the thorough review, Sam — really helpful. I've addressed all the points: split the identifier validation into separate format/check-digit rules with distinct codes, switched the invoice rule to direct field checks, and cleaned up the naming and translations.

PR #895, opened 2026-07-10
Extract included taxes from rate group sums with currency rounding

samlown:
## Summary

Fixes tax calculations when combining `prices_include` with the `currency` rounding rule. Previously each line's base was extracted and rounded individually, so with multiple lines the accumulated rounding drift meant the tax total's base plus amount no longer matched the document totals (e.g. 10 lines of 1.93 € at 21% VAT produced a base of 36.00 against a net total of 35.94).

With the currency rounding rule, included taxes are now determined from the sum of each rate group's tax-inclusive line totals, and shared back over the lines using cumulative rounding, so that:

- each rate's `base + amount` equals the sum of its original tax-inclusive line totals,
- `totals.total` equals the sum of the rate bases, and
- other categories — including retained taxes such as IRPF — accumulate their bases from the exact extracted line totals, without drift, even when spanning multiple rate groups.

The `precise` and `currency` rounding rules now have completely independent total calculators (`totals_calculator_precise.go` and `totals_calculator_currency.go`) so it is clear how each treats the data. The `precise` path is unchanged.

The Greek B2C example re-enables `prices_include`, which was commented out when the `round-then-sum` rule was originally introduced.

Note: a recalculation of a tax amount from its rounded base alone may occasionally differ by a single currency subunit from the amount derived from the tax-inclusive sums; this is inherent to any tax-inclusive scheme and is now mentioned in the rounding rule description.

## Testing

- New `tax` package tests covering multiple lines with a single rate, per-line rounding drift, retained taxes, retained taxes spanning multiple rate groups, and exempt lines.
- New `bill` package tests asserting full invoice totals with included VAT under currency rounding, with and without retained taxes.
- New PT example invoice (`examples/pt/invoice-pt-prices-include.yaml`) based on an anonymised customer document: 12 lines of 125.00 € with 6% VAT included and currency rounding, producing `base 1415.09 + tax 84.91 = 1500.00`.
- Regenerated Greek B2C example output: `base 6.81 + tax 1.63 = 8.44`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 - `tax`: the `currency` rounding rule combined with `prices_include` now determines each rate's tax amount from the sum of the tax-inclusive line totals and shares it back over the lines, so that bases, tax amounts, and document totals always add up, including when other categories such as retained taxes are present.
 
+### Fixed
+
+- `tax`: totals now reset the retained tax sum before recalculating, so repeated calculations over the same totals no longer accumulate stale retained amounts.
+
 ## [v0.502.2] - 2026-07-06
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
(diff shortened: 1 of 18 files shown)


cavalle (approved):
I haven't really double-checked the math, but the code looks great, and the aim is definitely sensible: the resulting tax rate totals shouldn't accumulate an error so large.

PR #899, opened 2026-07-15
en16931: validate party identity scope cardinality

alvarolivie:
## Summary

Adds two validation rules to the EN 16931 addon's `org.Party` identities, mirroring the standard's single-valued registration identifiers:

- **At most one** identity may carry the `legal` scope — the legal registration identifier (BT-30 seller / BT-47 buyer).
- **At most one** identity may carry the `tax` scope — the VAT identifier (BT-31 seller / BT-48 buyer).

Both rules share a small `orgIdentitiesScopeCount` helper. A party carrying zero such identities stays valid; only duplicates are rejected. `go generate .` regenerated `data/rules/eu-en16931.json` with the two new fault codes. Validation-only change — no normalization/schema output shift, and the examples suite passes unchanged, so no example document needed updating.

## Pre-Review Checklist


samlown (approved):

addons/eu/en16931/org.go
```diff
@@ -183,9 +183,43 @@ func orgPartyRules() *rules.Set {
...
+			n++
+		}
+	}
+	return n
+}
+
+func orgIdentitiesSingleLegalScope(val any) bool {
+	identities, ok := val.([]*org.Identity)
+	if !ok {
+		return true
```
samlown: This should be false by default, indicating that an incorrect value has been passed in.

samlown (changes requested):
Quick check on the boolean responses required.

samlown (approved):

PR #903, opened 2026-07-16
Normalize IT SDI customer to a single destination inbox

rilla:
## Context

FatturaPA's `DatiTrasmissione` block allows a single routing destination: per the specification, `PECDestinatario` may only be populated when `CodiceDestinatario` is the placeholder `0000000`. A GOBL customer carrying both an `it-sdi-code` and an `it-sdi-pec` inbox therefore describes a transmission that can never be expressed in valid FatturaPA — there is no correct way for a converter to write both.

## Changes

- When an invoice customer carries both inboxes, `it-sdi-v1` normalization now removes the PEC and keeps the code, so the document builds into deliverable FatturaPA. The code wins because it's the stronger routing signal — a PEC is only meaningful without one.
- Normalization was chosen over a validation error so that documents that build today keep building, and already-signed documents (which can't be re-normalized) don't start failing re-validation.
- The two IT examples that paired a code with a PEC now keep only the code, with golden outputs regenerated.

## Next

Nothing pending on this change.

https://claude.ai/code/session_01JDvrXj44FqsDPV4VSroccF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - `tax`: the `currency` rounding rule combined with `prices_include` now determines each rate's tax amount from the sum of the tax-inclusive line totals and shares it back over the lines, so that bases, tax amounts, and document totals always add up, including when other categories such as retained taxes are present.
 - `addons/eu/en16931`: party identities are now validated so that at most one may carry the `legal` scope (BT-30, BT-47) and at most one the `tax` scope (BT-31, BT-48).
 - `addons/it/sdi`: item attribute `type` is validated to be at most 10 characters, matching the FatturaPA `AltriDatiGestionali/TipoDato` (BT-160) field it maps to.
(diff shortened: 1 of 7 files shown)


alvarolivie (approved):
LGTM, some invoices may start failing if we've been accepting this. might make sense to normalize to code by removing the pec so as to not break others. (food for thought)

samlown (approved):
LGTM. I think this one also needs moving to its own gobl.it.sdi repo at some point.

rilla:
@alvarolivie good point, switched to normalization in 710381cd: when both inboxes are present, the PEC is dropped and the code kept. Went with normalization only (no validation backstop) — a validation rule would also fire on already-signed documents, which can't be re-normalized, so anything stored with both inboxes would start failing re-validation.

samlown:
@rilla Of course the other way to handle this is to just ignore one of the options in the conversion tool.

rilla:
@samlown I think you're right: this should go in the `gobl.fatturapa` conversor for two reasons:

1. It fixes already-signed documents. The add-on normalisation we implemented here would only run at build time, so a signed envelope with both inboxes would stay undeliverable. 
2. This is a FatturaPA serialization rule, so it belongs in the serializer. A customer genuinely can have both a PEC and a code; what's constrained is which one goes in the XML. The converter already owns exactly this kind of decision (e.g. the `0000000`/`XXXXXXX` defaults and routing rules are defined there too).

Closing this in favor of https://github.com/invopop/gobl.fatturapa/pull/77

PR #910, opened 2026-07-27
WIP: share removed included taxes out between the rows

cavalle:
Follow-up to #895, which fixed the tax-inclusive **totals** under `rounding: currency` but not what happens when those prices are subsequently **removed** — which is mandatory for SAF-T/AT, as the format cannot express VAT-inclusive prices. Portuguese payment receipts were still being rejected as a result (Pylon #5267, customer Cloudbeds).

The currency calculator extracts the included tax from the sum of each rate's tax-inclusive line totals and shares it back over the lines, so a rate's base is *not* the sum of the tax removed from each line on its own. `RemoveIncludedTaxes` did the latter, rounding every line down: 12 nights at a tax-inclusive 125.00 with 6% VAT came out with a base of 1415.04 instead of the 1415.09 the invoice was issued with, plus a `totals.rounding` of 0.06 that formats unable to express it silently drop.

- `tax`: new exported `TotalCalculator.ExtractIncludedTaxes`, returning the total of each taxable line with the included tax taken out. This is the same per-line distribution `calculateCurrencyTotals` already performed internally and threw away, so there is no second implementation to keep in step.
- `bill`: `RemoveIncludedTaxes` under `rounding: currency` now snapshots the net total every row should end up with *before* touching the prices, then nudges the net line prices (and fixed document discount/charge amounts) until they add up to it, recalculating in between so line discounts and charges settle. Each nudged price takes the least accuracy that still reproduces its line's sum, so the sums stay recalculable from the prices presented; large quantities pick up extra decimals as needed (`117.92453` for a quantity of 1000).
- Documents using `rounding: precise` are untouched — every pre-existing removal test passes unchanged.

Result for the invoice above: base **1415.09**, tax **84.91**, `total_with_tax` **1500.00**, no `totals.rounding`, with the lines carrying the remainder as 7 × 117.9245 + 5 × 117.93. Checked end-to-end against `at-pt`: SAF-T now renders `NetTotal` 1415.09 / `TaxPayable` 84.91 / `GrossTotal` 1500.00, and a receipt with base 1415.09 passes both `validatePaymentLine` and `AvailToPay.ApplyPayment` — the two checks that previously demanded different roundings of the same economics and so could not both be satisfied.

Multi-rate documents now get the correct base for every rate. A ±0.01 residual can still remain where a rate's tax cannot be recalculated from its own 2-decimal base at all (13% of 53.81 gives 7.00, but the tax-inclusive 60.80 implies 6.99); that is irreducible and still lands in `totals.rounding`, which is exactly what the field is for.

## Pre-Review Checklist

- [ ] Modified or created example GOBL documents to show my changes in use, if appropriate. — n/a: the example runner only calculates, it never calls `RemoveIncludedTaxes`, so no example can exercise this path. The `pt` and `es` `prices-include` examples added in #895 already cover the calculation side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Added
+
+- `tax`: `TotalCalculator.ExtractIncludedTaxes` provides the total of each taxable line with the included tax taken out, as shared by the `currency` rounding rule, so that the results add up to the bases determined by `Calculate`.
+
+### Fixed
+
+- `bill`: `RemoveIncludedTaxes` with the `currency` rounding rule now shares the removed tax out between the rows in the same way the tax totals do, instead of removing it from each price on its own. Line prices are nudged so that the resulting document keeps the tax bases it was issued with, and gains the accuracy needed for the line sums to still be recalculated from the prices presented. Previously a hotel invoice of 12 nights at a tax-inclusive 125.00 with 6% VAT ended up with a base of 1415.04 instead of 1415.09, and a `totals.rounding` of 0.06 that formats unable to express it would silently drop.
+
 ## [v0.503.0] - 2026-07-15
 
 ### Added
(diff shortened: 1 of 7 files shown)


samlown:
This one I think is fixed by #923, closing for the time being.

PR #915, opened 2026-07-31
Add fi-finvoice-v3 addon for Finnish Finvoice 3.0

rilla:
## Context

Finvoice 3.0 is the Finnish national e-invoicing format, conformant with EN 16931. Several elements that are optional in EN 16931 are mandatory in the Finvoice schema — most notably the `EpiDetails` payment block, which every invoice must carry, credit notes included.

This adds a `fi-finvoice-v3` addon following the same split as `de-xrechnung`: it declares `Requires: eu-en16931-v2017` and only adds the Finvoice-specific tightening on top.

## Changes

- **New `addons/fi/finvoice` package** defining `fi-finvoice-v3`, wired into `addons/addons.go` with regenerated data files.
- **Customer must be present and named** — `BuyerOrganisationName` is schema-mandatory.
- **Payment details are required on every invoice** — `EpiDetails` is mandatory even on credit notes, so the rule doesn't inherit EN 16931's "only when an amount is due" condition (BR-CO-25).
- **Credit transfer instructions with an IBAN** — the instructions key must be `credit-transfer` (extensions like `credit-transfer+sepa` are fine) and the first credit transfer entry must carry an IBAN (`EpiAccountID`), validated with the ISO 7064 mod 97-10 checksum. Country-specific lengths are not checked.
- **Payment reference is required** (`EpiReference`) — presence only; Finnish reference / RF check-digit validation could come later.
- **At least one dated due date** (`EpiDateOptionDate`) — EN 16931 accepts terms with notes only.
- **Normalizers strip grouping spaces** from IBANs (also upper-casing them) and payment references, so display-formatted values like `FI21 1234 5600 0007 85` validate.
- **BIC stays recommend-only** — `EpiBfiIdentifier` is optional in the schema.

Sources: [Finvoice 3.0 XSD](https://file.finanssiala.fi/finvoice/css/270923/Finvoice3.0.xsd) and the [Finvoice implementation guidelines](https://www.finanssiala.fi/en/topics/finvoice-implementation-guidelines/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Added
+
+- `addons/fi/finvoice`: new `fi-finvoice-v3` addon for the Finnish Finvoice 3.0 format. Builds on `eu-en16931-v2017` and adds the Finvoice-specific requirements: a named customer, and unconditional payment details — credit transfer instructions with a checksum-validated IBAN, a payment reference, and at least one due date — since Finvoice's `EpiDetails` block is mandatory on every invoice, including credit notes. IBANs and payment references are normalized to their machine form (grouping spaces removed, IBANs upper-cased).
+
 ## [v0.503.0] - 2026-07-15
 
 ### Added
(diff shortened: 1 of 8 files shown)


rilla:
Closing: new addons are moving out of the core module following the external-addon model (#847), so this one will land in its own module instead.

PR #917, opened 2026-08-03
Add IsIBAN and IsBIC rules tests to pay

rilla:
## Context

Bank-detail validation keeps coming up in addon rule sets, and `rules/README.md` asks for tests that are generally useful to a GOBL type to live in that type's package rather than privately in an addon. These two are the IBAN and BIC checks the Finnish Finvoice addon needs.

## Changes

- **`pay.IsIBAN`** — the ISO 13616 structure composed with its ISO 7064 mod 97-10 check digits. Per-country lengths are not enforced.
- **`pay.IsBIC`** — the ISO 9362 structure.
- Both live in `pay/validation.go`, following `num` and `uuid`. Neither hand-checks absent values: the pattern tests in `rules/is` skip them so that presence is asserted separately, and an empty IBAN is invalid.
- **`is.AllOf`** — composes several tests into one that passes only when all of them do, alongside the existing `AnyOf` and `OneOf` and built the same way. It is what lets `IsIBAN` read as its two conditions. `is.MatchesRegexp` rather than `is.Matches` inside it, since composites do not compile their nested tests.
- `rules/README.md` gains a `pay` section in the domain-tests tables, including the pre-existing `HasValidMeansKey`, plus the `AllOf` entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Added
 
+- `pay`: `IsIBAN` and `IsBIC` rules tests for bank details — the ISO 13616 structure with its ISO 7064 mod 97-10 check digits, and the ISO 9362 structure. Available to rule sets that need them; no core rule applies them.
+- `rules/is`: `AllOf` composes several tests into one that passes only when all of them do, alongside the existing `AnyOf` and `OneOf`.
 - `net`: request tokens (README §5.5): short-lived ES256 JWTs sent as `Authorization: Bearer` on `/who` and `/inbox` requests, identifying the requester — possibly a trusted intermediary distinct from the envelope's signer. `net.NewToken` mints one, `Client.VerifyToken` / `Client.VerifyAuthorization` verify inbound tokens against the issuer's published key, audience, and a 30s–5m freshness window, and the `net.WithIdentity` client option attaches one to every who and inbox request. New `ErrTokenInvalid` and `ErrTokenExpired` sentinels.
 - `net`: `Client.Send` delivers a signed envelope to an address's inbox: 202 is success, other non-retryable 4xx report `ErrInboxRejected`. The `Poster` interface extends a `Fetcher` with POST support; `HTTPFetcher` implements it.
 - `net`: `/who` may answer `202 Accepted` (new `ErrPending` sentinel) to record the authenticated request for deferred disclosure: the owner decides per requester and, if approved, delivers its party envelope to the requester's inbox later.
--- a/pay/validation.go
+++ b/pay/validation.go
@@ -1,0 +1,47 @@
+package pay
+
+import (
+	"regexp"
+
+	"github.com/invopop/gobl/rules/is"
+)
+
+var (
+	// ibanPattern reflects the ISO 13616 structure: country code, check
+	// digits, and 11 to 30 characters of national account data. Per-country
+	// lengths are not enforced.
+	ibanPattern = regexp.MustCompile(`^[A-Z]{2}\d{2}[A-Z0-9]{11,30}$`)
+
+	// bicPattern reflects the ISO 9362 structure: institution and country
+	// codes, a location code, and an optional branch code.
+	bicPattern = regexp.MustCompile(`^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?$`)
+)
+
+var (
+	// IsIBAN confirms the value is an International Bank Account Number.
+	IsIBAN = is.AllOf(
+		is.MatchesRegexp(ibanPattern),
+		is.StringFunc("has matching check digits", ibanCheckDigitsMatch),
+	)
+
+	// IsBIC confirms the value is a Business Identifier Code.
+	IsBIC = is.MatchesRegexp(bicPattern)
+)
+
+// ibanCheckDigitsMatch runs the ISO 7064 mod 97-10 calculation over the
+// account number moved in front of the country code and check digits, with
+// letters converted to two digits each (A is 10, Z is 35).
+func ibanCheckDigitsMatch(iban string) bool {
+	if len(iban) < 5 {
+		return false // too short to carry check digits and an account number
+	}
+	n := 0
+	for _, r := range iban[4:] + iban[:4] {
+		if r >= 'A' && r <= 'Z' {
+			n = (n*100 + int(r-'A') + 10) % 97
+		} else {
+			n = (n*10 + int(r-'0')) % 97
+		}
+	}
+	return n == 1
+}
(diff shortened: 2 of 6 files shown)


rules/README.md
```diff
@@ -258,6 +258,14 @@ Tax — `github.com/invopop/gobl/tax`:
 | `tax.HasAddon(key)` / `tax.AddonIn(keys...)`  | Object declares the addon / context carries it              |
 | `tax.TagsIn(tags...)`                         | All of the object's tags are in the list                    |
 
+Payments — `github.com/invopop/gobl/pay`:
+
+| Test                   | Notes                                                          |
+| ---------------------- | -------------------------------------------------------------- |
+| `pay.HasValidMeansKey` | Instructions key is based on one of the defined means keys     |
+| `pay.ValidIBAN`        | ISO 13616 shape + ISO 7064 mod-97 checksum; empty values pass  |
```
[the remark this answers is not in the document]
rilla: Aligned to "mod 97-10".

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Added
 
+- `pay`: `ValidIBAN` and `ValidBIC` rules tests — the ISO 13616 shape with the ISO 7064 mod-97 checksum, and the ISO 9362 shape respectively. Available for rule sets that need to validate bank details; nothing in core applies them by default. Empty values pass, so presence is asserted separately.
```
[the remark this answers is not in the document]
rilla: Aligned to "mod 97-10".

pay/bic.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// bicPattern is the ISO 9362 BIC shape: bank and country codes plus a
+// location suffix, with an optional branch code.
+var bicPattern = regexp.MustCompile(`^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?$`)
+
+// ValidBIC is a rules test that checks a string is a well-formed BIC
+// according to the ISO 9362 shape. Empty values pass, so presence must be
+// asserted separately.
+var ValidBIC = is.MatchesRegexp(bicPattern)
```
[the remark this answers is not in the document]
rilla: Switched to `is.StringFunc("valid BIC", ...)` — confirmed the raw regex was landing in the exported rules JSON (`data/rules/*.json` shows `"tests": "matches ^..."` for other MatchTest uses), so the stable description is the right call for an exported test.

pay/bic.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// bicPattern is the ISO 9362 BIC shape: bank and country codes plus a
+// location suffix, with an optional branch code.
+var bicPattern = regexp.MustCompile(`^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?$`)
+
+// ValidBIC is a rules test that checks a string is a well-formed BIC
+// according to the ISO 9362 shape. Empty values pass, so presence must be
+// asserted separately.
+var ValidBIC = is.StringFunc("valid BIC", validBIC)
```
samlown: Is "Valid" the correct prefix? Probably should be `pay.IsBIC`? Also, you shouldn't be empty checking, empty means wrong by default, so it'd be best to use the is.Matches method.

pay/iban.go
```diff
@@ -0,0 +1,35 @@
...
+
+// ibanPattern covers the ISO 13616 shape: country code, two check digits,
+// and 11 to 30 alphanumeric BBAN characters, upper case and without
+// grouping spaces. Exact per-country lengths are not checked.
+var ibanPattern = regexp.MustCompile(`^[A-Z]{2}\d{2}[A-Z0-9]{11,30}$`)
+
+// ValidIBAN is a rules test that checks a string is a well-formed IBAN:
+// the ISO 13616 shape and the ISO 7064 mod 97-10 checksum. Empty values
+// pass, so presence must be asserted separately.
+var ValidIBAN = is.StringFunc("valid IBAN", validIBAN)
```
samlown: Same here, I'd probably go for `IsIBAN`.

pay/iban.go
```diff
@@ -0,0 +1,35 @@
...
+
+func validIBAN(iban string) bool {
+	if iban == "" {
+		return true
+	}
+	if !ibanPattern.MatchString(iban) {
```
samlown: You could also use `is.AllOf` with two conditions to make the rule more readable.

pay/iban.go
```diff
@@ -0,0 +1,35 @@
...
+var ibanPattern = regexp.MustCompile(`^[A-Z]{2}\d{2}[A-Z0-9]{11,30}$`)
+
+// ValidIBAN is a rules test that checks a string is a well-formed IBAN:
+// the ISO 13616 shape and the ISO 7064 mod 97-10 checksum. Empty values
+// pass, so presence must be asserted separately.
+var ValidIBAN = is.StringFunc("valid IBAN", validIBAN)
+
+func validIBAN(iban string) bool {
+	if iban == "" {
+		return true
```
samlown: Pretty sure an empty IBAN is not a valid IBAN :-)

samlown (changes requested):
Naming here is a bit inconsistent and the rules/README.md doesn't appear to have been followed. Out of interest, which model did you use here?

pay/validation.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// bicPattern reflects the ISO 9362 structure: institution and country
+	// codes, a location code, and an optional branch code.
+	bicPattern = regexp.MustCompile(`^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?$`)
+)
+
+var (
+	// IsIBAN confirms the value is an International Bank Account Number
+	// whose ISO 7064 mod 97-10 check digits match.
+	IsIBAN = is.StringFunc("is valid IBAN", isIBAN)
```
samlown: Does this work?

```suggestion
	IsIBAN = is.AllOf(
	    is.Matches(iban),
	    ibanCheckDigitsMatch(iban), // this would need to return a rules.Test
	)
```

rilla: I think it wouldn't right now because looks like `is.AllOf` doesn't exist yet, but if we added it following `is.AnyOf` and `is.OneOf`, we could then use it with something like this:

``` go
  var (
        // ibanPattern reflects the ISO 13616 structure: country code, check
        // digits, and 11 to 30 characters of national account data. Per-country
        // lengths are not enforced.
        ibanPattern = regexp.MustCompile(`^[A-Z]{2}\d{2}[A-Z0-9]{11,30}$`)
        ...
  )

  var (
        // IsIBAN confirms the value is an International Bank Account Number.
        IsIBAN = is.AllOf(
                is.MatchesRegexp(ibanPattern),
                is.StringFunc("has matching check digits", ibanCheckDigitsMatch),
        )

        // IsBIC confirms the value is a Business Identifier Code.
        IsBIC = is.MatchesRegexp(bicPattern)
  )
```

The only significant change from your snippet would be using `is.MatchesRegexp` instead of `is.Matches`: composites don't compile their nested tests, so a lazy `is.Matches` inside one would panic. 

If this looks good, happy to add `is.AllOf` as part of this PR and update the code.

samlown (approved):
Quite like `AllOf`! nice.

PR #922, opened 2026-08-06
Add Slovakia (SK) tax regime

Liuck27:
## Summary

Adds the Slovak (SK) tax regime, structured on `regimes/fi`.

- **VAT** with the three rates in force since the 2025 reform: standard 23%, reduced 19%, super-reduced 5% (all since 2025-01-01), plus the prior 20% standard rate (since 2011-01-01) so earlier invoices still resolve. Zero, exempt, reverse-charge, export and outside-scope keys come from `tax.GlobalVATKeys()`, as in other regimes.
- **IČ DPH tax identity**: normalization strips the `SK` prefix, spaces and hyphens; validation requires 10 digits with a non-zero first digit, the whole number divisible by 11.
- **EUR** currency, **`Europe/Bratislava`** time zone, Slovak (`sk`) translations alongside English.
- **Example invoice** exercising all three rates, with items matching each rate's documented scope (software services, electricity, books).
- Tests, generated regime/rules data, CHANGELOG.

## Decisions

**Rate history covers the current structure plus the prior standard rate only.** Before 2025 there was a single 10% reduced rate, covering the goods listed in annex 7 of the VAT act (medicines, books). The 2025 reform abolished that tier rather than renaming it, introducing 19% over a different set of goods, while the old 10% scope maps more closely to today's 5%. Attaching that history to either current rate would misstate which goods were taxed at which rate, so it is left out. Invoices dated before 2025-01-01 using `reduced` or `super-reduced` therefore fail with `rate value unavailable`; the standard rate resolves back to 2011. Several regimes have the same shape: `fr` (reduced tiers only from 2014), `se` (from 1996), `ie` (general only from 2012 while reduced reaches 2003).

**Why the 20% rate cites two laws.** Slovakia's VAT act (no. 222/2004) normally states its rates in section 27, but the 20% rate did not start there. A 2010 amendment (no. 490/2010) introduced it in section 85j instead, as a temporary measure due to expire once the national deficit fell below 3%. So between 2011 and 2014, section 27 still read 19% while the rate actually charged was 20%. The deficit condition was never met, and from 2015 the 20% was written into section 27 itself, where it stayed until the 2025 reform. The rate therefore ran unbroken at 20% from 2011 to 2024, and only the part of the law holding it changed. Citing section 27 alone would not support the 2011 start date, so both laws are listed as sources.

**Tax identity validation stops short of the third-digit rule.** The reference implementation also restricts the third digit to `[2346-9]`. That is not adopted here: sources disagree on it, the algorithm is not published by the tax authority, and wrongly rejecting a valid VAT number blocks a real business from invoicing whereas wrongly accepting one is only a data-quality issue. The non-zero first digit is adopted, since without it `0000000000` validates (`0 % 11 == 0`).

## Deliberately out of scope

Regime only: no addon, no format conversion, no tax-authority transmission, no invoice chaining. Slovakia's e-invoicing requirements belong in an addon rather than the regime. There are also no supplier-specific invoice rules and no requirement for a tax ID on invoices, since Slovakia has a VAT-registration threshold below which a business has none and the regime resolves from `tax_id.country` alone.

## Open question

The tax authority does not publish the IČ DPH check algorithm. The modulo-11 rule is consistently documented across independent validators and holds for every valid identifier tested, but it is not an official specification. This is flagged in a code comment, following the precedent in `regimes/sg`.

## How to test

```bash
go generate .
go test ./...
go test ./regimes/sk/ -cover   # 100%
```

## Sources

- [Slov-Lex, zákon č. 222/2004 Z. z.](https://www.slov-lex.sk/ezbierky/pravne-predpisy/SK/ZZ/2004/222/), § 27 sets the current rates
- [Slov-Lex, zákon č. 490/2010 Z. z.](https://www.slov-lex.sk/ezbierky/pravne-predpisy/SK/ZZ/2010/490/), introduced the 20% standard rate
- [Finančná správa, daň z pridanej hodnoty](https://www.financnasprava.sk/sk/podnikatelia/dane/dan-z-pridanej-hodnoty) and its [English portal](https://www.financnasprava.sk/en/)
- Checksum (not authority-published): [`ltns35/go-vat`](https://github.com/ltns35/go-vat/blob/main/countries/slovakia.go), already cited by `regimes/de` and `regimes/dk`, and [VatDB](https://vatdb.com/guides/validate-sk-vat-number/)

All sources are also recorded in code as `cbc.Source` entries with retrieval dates.

## Pre-Review Checklist

### DIFF
--- a/regimes/sk/tax_identity.go
+++ b/regimes/sk/tax_identity.go
@@ -1,0 +1,61 @@
+package sk
+
+import (
+	"regexp"
+
+	"github.com/invopop/gobl/cbc"
+	"github.com/invopop/gobl/rules"
+	"github.com/invopop/gobl/rules/is"
+	"github.com/invopop/gobl/tax"
+)
+
+// taxCodeRegexp matches a normalized Slovak IČ DPH: exactly 10 digits, the first
+// of which is non-zero. The "SK" country prefix, spaces and hyphens are removed
+// during normalization (tax.NormalizeIdentity), so only the digits reach this
+// validation.
+var taxCodeRegexp = regexp.MustCompile(`^[1-9]\d{9}$`)
+
+func taxIdentityRules() *rules.Set {
+	return rules.For(new(tax.Identity),
+		rules.When(tax.IdentityIn(CountryCode),
+			rules.Field("code",
+				rules.AssertIfPresent("01", "tax identity code for SK must be 10 digits starting with 1-9 and divisible by 11",
+					is.Func("valid", validateTaxCode),
+				),
+			),
+		),
+	)
+}
+
+func validateTaxCode(value any) bool {
+	code, ok := value.(cbc.Code)
+	if !ok {
+		return false
+	}
+	val := code.String()
+
+	if !taxCodeRegexp.MatchString(val) {
+		return false
+	}
+	return validateTaxCodeChecksum(val)
+}
+
+// Slovak IČ DPH checksum: the whole 10-digit number must be divisible by 11.
+//
+// A 10-digit number can reach ~10^10, which overflows a 32-bit int, so the digits
+// are accumulated into an int64. Digit conversion via val[i]-'0' assumes the input
+// contains only ASCII digits, which is guaranteed by the regex in validateTaxCode.
+//
+// Note: the Slovak tax authority does not publish this algorithm; the modulo-11
+// rule is consistently documented by EU VAT-number validators and matches all
+// known valid identifiers (e.g. SK2020273893).
+//
+// Reference: https://github.com/ltns35/go-vat/blob/main/countries/slovakia.go
+// Reference: https://vatdb.com/guides/validate-sk-vat-number/
+func validateTaxCodeChecksum(val string) bool {
+	var n int64
+	for i := range len(val) {
+		n = n*10 + int64(val[i]-'0')
+	}
+	return n%11 == 0
+}
--- a/regimes/sk/tax_identity.go
+++ b/regimes/sk/tax_identity.go
@@ -2,8 +2,8 @@ package sk
 
 import (
 	"regexp"
+	"strconv"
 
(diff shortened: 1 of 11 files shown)


regimes/sk/tax_identity.go
```diff
@@ -0,0 +1,61 @@
...
+// during normalization (tax.NormalizeIdentity), so only the digits reach this
+// validation.
+var taxCodeRegexp = regexp.MustCompile(`^[1-9]\d{9}$`)
+
+func taxIdentityRules() *rules.Set {
+	return rules.For(new(tax.Identity),
+		rules.When(tax.IdentityIn(CountryCode),
+			rules.Field("code",
+				rules.AssertIfPresent("01", "tax identity code for SK must be 10 digits starting with 1-9 and divisible by 11",
+					is.Func("valid", validateTaxCode),
```
samlown: Preference here is to split into specific rules rather than a `is.Func` catch all. For example, there could be a specific assertion around the pattern matching of the string, and a second for the checksum.

samlown (changes requested):
Looks good thanks! I'd recommend changing the tax ID validations to more specific rules.

regimes/sk/tax_identity.go
Liuck27: Done. Split into two assertions: IDENTITY-01 for the 10-digit pattern and IDENTITY-02 for the modulo-11 checksum, each with its own message. Since assertions are evaluated independently, the checksum no longer depends on the pattern rule having run first, so it now parses with strconv.ParseInt and rejects non-digit input on its own. Dropping the is.Func wrapper also removed the cbc.Code type assertion, which took the package to 100% coverage.

samlown: This is great! thanks.

samlown (approved):
LGTM. Thanks!

PR #938, opened 2026-08-26
Malta (MT) VAT tax regime is implemented

AlfonsoUceda:
## Summary

This adds a regime-only VAT (Value Added Tax) definition for Malta, built on the same patterns as the recent EU (European Union) regimes (`ie`, `fi`).

I picked Malta partly because it is uncontested and partly because it publishes its VAT law in English, so I could source every rule from the primary text — the Value Added Tax Act, Chapter 406 (https://legislation.mt/eli/cap/406/eng) — instead of a translation.

The rate, exemption, invoice-content, correction and VAT-number rules all trace back to that Act (Eighth and Fifth Schedules, art. 50 and the Twelfth Schedule, Tenth Schedule, and art. 13 respectively), with the current rates cross-checked against the Malta Tax and Customs Administration (MTCA) "VAT Rates & Exemptions" guidance.

## How VAT works in Malta

VAT runs at a standard 18 %, with reduced rates of 12 %, 7 % and 5 % on specific goods and services.

What is unusual is that Malta splits its VAT-free supplies into two kinds: "exempt with credit", where no VAT is charged but the supplier still reclaims its input VAT (economically a zero rate), and "exempt without credit", where no VAT is charged and nothing is reclaimed.

Businesses are identified by an `MT` + 8-digit VAT number, invoices are corrected with credit and debit notes, and the currency is the euro.

## The main modelling decision

The decision I'd most like a second opinion on is how those two kinds of exemption map onto GOBL's keys.

"Exempt with credit" (Fifth Schedule Part One — exports, food, pharmaceuticals) is really zero-rating: no VAT is charged, but input VAT stays recoverable, so I mapped it to `zero`.
"Exempt without credit" (Part Two — finance, insurance, residential letting) maps to `exempt`.

Collapsing both into `exempt` would drop the input-credit distinction the law makes, and since GOBL's keys don't encode recoverability themselves, this key choice carries the whole distinction.
The `invoice-mt-zero-rated` and `invoice-mt-exempt` examples show the two producing different combos.

## Smaller decisions

- I validate only the supplier's VAT number, not the customer's. The Twelfth Schedule requires the customer's number on a full tax invoice, but simplified invoices (≤ €100) and business-to-consumer (B2C) receipts legitimately omit it — enforcing it properly means modelling the €100 threshold and B2C detection, which felt like more than this first PR should carry. I documented the requirement and left it as an open question.

- The 18/12/7/5 % rates use `general`/`reduced`/`super-reduced`/`special`, following `ie`. Those keys are just identifiers, so I'm happy to rename to whatever convention you prefer.

- The VAT-number checksum is a mod-37 scheme that the MTCA and the EU VAT Information Exchange System (VIES) don't actually publish — they document the `MT` + 8-digit format but not the algorithm. I took it from community implementations, labelled it non-official in the code, and pinned it down with tests against real-shaped numbers plus the mod-0 edge case. I'd rather ship a labelled best-effort check than one that looks authoritative but isn't.

## Deliberately out of scope

- A Peppol / EN 16931 (the EU e-invoicing standard) addon — Malta only mandates business-to-government (B2G) receive-and-process, so this belongs in a later `addons/mt/…` that layers on with no regime change.

- The Malta Business Registry (MBR) company number and ID-card numbers — not the invoice tax identity.

- Article 11 small-undertaking numbers — explicitly not VAT numbers.

- Fiscal receipts, One Stop Shop (OSS), and margin schemes.

## Testing

Unit tests cover the failure and edge cases the example harness can't (bad checksums, normalization, the supplier rule).

The six annotated examples double as integration tests — the harness calculates and validates each against a committed golden, so the golden diffs are the real proof the tax computes correctly.

## Open questions

1. Customer VAT number — enforce it conditionally, or is documenting it enough for now?
2. Reduced-rate key names — keep `reduced`/`super-reduced`/`special`, or match a convention I've missed?
3. The non-official mod-37 checksum — ship it labelled, or prefer format-only validation?
4. Standard-rate history — keep the 1999 → 2004 (15 % → 18 %) step, given the 2004 date is only non-officially sourced?


regimes/mt/mt.go
```diff
@@ -0,0 +1,89 @@
...
+			i18n.EN: "Malta",
+			i18n.MT: "Malta",
+		},
+		Description: i18n.String{
+			i18n.EN: here.Doc(`
+				Malta's VAT is governed by the Value Added Tax Act (Cap. 406) and administered
+				by the Malta Tax and Customs Administration (MTCA). As an EU member state Malta
+				follows the EU VAT Directive (2006/112/EC).
+
+				The standard rate is 18%, with reduced rates of 12%, 7% and 5%. Supplies that are
```
samlown: Rate values should not be included, I'd keep these simple.

regimes/mt/tax_identity.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return validateTaxCode(code) == nil
+}
+
+// validateTaxCode validates a normalized Maltese VAT identity code: 8 digits
+// with a valid check number.
+func validateTaxCode(code cbc.Code) error {
+	val := code.String()
+	if !taxCodeRegexp.MatchString(val) {
+		return errors.New("invalid format")
```
samlown: I'd avoid returning internal errors that are never actually used elsewhere.

regimes/mt/tax_identity.go
```diff
@@ -0,0 +1,72 @@
...
+// small-undertaking numbers have no prefix and are not VAT identification
+// numbers (Cap. 406 art. 13), so they are not validated here.
+var taxCodeRegexp = regexp.MustCompile(`^\d{8}$`)
+
+func taxIdentityRules() *rules.Set {
+	return rules.For(new(tax.Identity),
+		rules.When(tax.IdentityIn(CountryCode),
+			rules.Field("code",
+				rules.AssertIfPresent("01", "invalid Maltese VAT identity code",
+					is.Func("valid", isValidTaxIdentityCode),
```
samlown: Most likely we need a regexp comparison first, followed with the checksum test. Each of these can be tested an reported independently.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Added
 
+- `regimes/mt`: added the Malta (`MT`) VAT regime — standard 18% rate with 12%/7%/5%
+  reduced rates, exempt-with-credit mapped to `zero` and exempt-without-credit to `exempt`,
+  mod-37 VAT-number validation, and credit-note/debit-note corrections.
```
samlown: We don't need so much detail here.

samlown (changes requested):
Examples are good. I'd refined some of the texts and include additional assertions inside the tax identity rules.

AlfonsoUceda:
Hey @samlown I've addressed your and copilot comments ;)

regimes/mt/bill_invoices.go
```diff
@@ -0,0 +1,24 @@
...
+
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(CountryCode)),
+			rules.Field("supplier",
+				// Cap. 406 art. 50(5) and Twelfth Schedule item 3(c): a tax invoice
+				// must show the supplier's VAT identification number.
+				rules.Assert("01", "invoice supplier in Malta must have a VAT tax ID code",
+					org.PartyHasTaxIDCode(),
```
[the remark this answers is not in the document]
AlfonsoUceda: Agreed that a reverse-charge invoice needs the customer's VAT number (Twelfth Schedule 3(d)). This is the flip side of the same layering: customer-side and reverse-charge business rules are enforced by the EN 16931 addon, and no base EU regime validates the customer party — they only assert that the supplier is identifiable. I scoped the regime to the supplier VAT-ID requirement on purpose and documented the customer requirement, raising the general "should the regime validate the customer VAT number" as an open question, because doing it properly means also modelling the simplified (≤ €100) and B2C carve-outs. If you'd prefer a narrow, reverse-charge-only customer-VAT assertion in the regime itself, I'm happy to add it — just say the word.

regimes/mt/bill_invoices.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func billInvoiceRules() *rules.Set {
+	return rules.For(new(bill.Invoice),
+		rules.When(
+			is.InContext(tax.RegimeIn(CountryCode)),
+			rules.Field("supplier",
+				// Cap. 406 art. 50(5) and Twelfth Schedule item 3(c): a tax invoice
+				// must show the supplier's VAT identification number.
+				rules.Assert("01", "invoice supplier in Malta must have a VAT tax ID code",
```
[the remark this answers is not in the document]
AlfonsoUceda: Good catch on the legal point — the Twelfth Schedule (and Directive 2006/112/EC art. 226(11)) do require an exempt invoice to reference the applicable provision. In GOBL, though, that reference is an e-invoicing concern handled by the EN 16931 addon rather than the tax regime: addons/eu/en16931 already requires exempt categories to carry either a CEF VATEX code or an exemption note in tax.notes. None of the base EU regimes (ie, fi, se, nl, es) enforce exemption notes themselves — they leave it to the addon. Adding it here would duplicate that rule and make Malta inconsistent with the others, so I've kept the regime to the minimum the law needs on its own and left the exemption reference to the addon layer (noted as out of scope in the PR description, with the addon attachment seam).

PR #939, opened 2026-08-26
Add SDI issuer type extension for customer and third-party issuance

rilla:
Adds the `it-sdi-issuer-type` extension so an Italian invoice can say who compiled it when that was not the supplier. FatturaPA carries this in `SoggettoEmittente` (issuing party), with `TZ` for a third party and `CC` for the customer (*cessionario/committente*). The field is required whenever the compiler differs from the supplier (art. 21 DPR 633/1972), and until now GOBL had no way to express it. This follows `es-verifactu-issuer-type`, which covers the same concept for Spain.

## Changes

- **Adds `it-sdi-issuer-type`** with values `TZ` and `CC`, sourced from the AdE compilation guide.
- **Sets `TZ` automatically** when the invoice's ordering names an `issuer`, as the Spanish addon does.
- **Requires the issuer to be identifiable.** An `ordering.issuer` must carry a tax ID code or a fiscal code identity, since the FatturaPA third-party block needs one of the two.
- **Sets `CC` only when the parties are distinct.** A `self-billed` invoice gets `CC` when the customer's tax ID differs from the supplier's. Italy also uses self-billing for documents where supplier and customer are the same company and the supplier files (TD27 is the clearest case) — there the field stays unset, and integrators can still set it explicitly.
- **Adds an example invoice** issued by a third party.

## Pre-Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 - `net`: added `SandboxAuthorities` (defaulting to `lookup.sandbox.gobl.org`)
   and `WithSandbox`. Sandbox and live trust lists remain separate.
+- `it-sdi`: `it-sdi-issuer-type` extension for the FatturaPA `SoggettoEmittente`
+  field, set to `TZ` when the invoice ordering names an `issuer` and to `CC` on
+  self-billed invoices whose customer tax ID differs from the supplier's.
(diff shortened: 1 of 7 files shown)


rilla:
Retargeted: the it-sdi addon is moving out of core (#876), so this extension now lands in the extracted addon instead — invopop/gobl.fatturapa#82, with the converter mapping in invopop/gobl.fatturapa#81.

PR #940, opened 2026-08-26
Describe the it-sdi addon and how Italian parties are identified

rilla:
The `it-sdi-v1` addon had no description, so its page on docs.gobl.org opened straight into scenarios with nothing on what the addon is for or how an Italian party is identified.

Adds one, in the style of the other addons: what the addon does, which identifier and inbox each kind of customer carries, and the placeholder values FatturaPA expects where a party has no code of its own — which the conversion fills in, so they should never be written by hand.

`data/addons/it-sdi-v1.json` regenerated with `go generate .`.


rilla:
Closing — the description will be added directly on the extraction branch (invopop/gobl.fatturapa#71) so it lands with the addon in its new home.

PR #945, opened 2026-08-29
Add Japan (JP) tax regime

fenomenog4:
Adds a new tax regime for **Japan (JP)**.

- New `regimes/jp` regime: currency `JPY`, time zone `Asia/Tokyo`, VAT tax scheme.
- **Consumption Tax (消費税)** modelled on the shared VAT category — 10% standard rate and 8% reduced rate (food and drink excluding alcohol and dining out, and qualifying newspaper subscriptions), with historical rates of 3% (1989), 5% (1997) and 8% (2014).
- **Qualified-invoice registration number** validation: `T` + 13 digits, enforcing the NTA mod-9 check digit documented for the Corporate Number (法人番号). The check digit is the first of the 13 digits (prepended) and is verified against the NTA's worked example.

**Design notes**
- Why Japan: it's one of the world's largest economies ([wikipedia](https://en.wikipedia.org/wiki/Economy_of_Japan)), its Consumption Tax is a credit-invoice VAT that maps cleanly onto GOBL's existing VAT model. It already runs JP PINT — a structured e-invoicing specification built on Peppol PINT — giving a clear path to a future JP PINT addon.
- No unconditional supplier-tax-ID invoice rule: Japan has legitimate non-registered issuers (a registration number is only mandatory for an invoice functioning as a *qualified* invoice), and GOBL core can't distinguish the two from the document alone — so requiring one would reject valid non-qualified invoices.
- No rule forces a supplier tax ID. Only registered businesses have a registration number, and it's only required when the invoice needs to qualify. The invoice itself doesn't say which case it is, so requiring one would reject valid non-qualified invoices.
- The check digit is enforced on every `T` number. Note that numbers the NTA issues to registrants without a Corporate Number are not guaranteed by any published spec to satisfy the same algorithm; this is an accepted trade-off in favour of stronger validation. The assumption can be validated against real registration numbers during commissioning tests, and relaxed to format-only if it turns out not to hold.

**Sources**
- [National Tax Agency — Information about Consumption Tax](https://www.nta.go.jp/english/taxes/consumption_tax/index.htm)
- [NTA — Outline of the invoice system (PDF)](https://www.nta.go.jp/english/taxes/consumption_tax/pdf/2022/general_17.pdf)
- [NTA — Registration number format](https://www.invoice-kohyo.nta.go.jp/about-toroku/index.html)
- [NTA — Corporate Number check-digit calculation (PDF)](https://www.houjin-bangou.nta.go.jp/documents/checkdigit.pdf) — Japanese only; a workable English translation can be obtained via [DeepL's file translator](https://www.deepl.com/es/translator/files) if you can't read Japanese :)
- [LookupTax — Japan Corporate Number validator](https://lookuptax.com/validate/japan/corporate-number) third-party check-digit validation

## Pre-Review Checklist


examples/jp/invoice.json
samlown, line 1: I think it'd be useful to have more examples here.

regimes/jp/jp.go
```diff
@@ -0,0 +1,70 @@
...
+		norm.When(tax.IdentityIn(CountryCode), norm.For(normalizeTaxIdentity)),
+	)
+}
+
+// New provides the tax region definition for Japan.
+func New() *tax.RegimeDef {
+	return &tax.RegimeDef{
+		Country:   CountryCode,
+		Currency:  currency.JPY,
+		TaxScheme: tax.CategoryVAT,
```
samlown, line 32: Are you sure it is VAT? Information I've found suggests it has a different name, like `JCT`.

regimes/jp/jp.go
```diff
@@ -0,0 +1,70 @@
...
+				Title: i18n.NewString("National Tax Agency - Outline of the invoice system"),
+				URL:   "https://www.nta.go.jp/english/taxes/consumption_tax/pdf/2022/general_17.pdf",
+			},
+		},
+		TimeZone: "Asia/Tokyo",
+		Description: i18n.String{
+			i18n.EN: here.Doc(`
+				Japan levies a Consumption Tax (消費税, shōhizei) administered by the
+				National Tax Agency (NTA). It is a value-added tax charged at a standard
+				rate of 10% with a reduced rate of 8% for food and drink (excluding alcohol
```
samlown: I'd avoid mentioning percentages directly as these are also reflected in the tax_categories and implies two places would need to be updated.

regimes/jp/tax_categories.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/invopop/gobl/num"
+	"github.com/invopop/gobl/pkg/here"
+	"github.com/invopop/gobl/tax"
+)
+
+func taxCategories() []*tax.CategoryDef {
+	return []*tax.CategoryDef{
+		// Consumption Tax (modelled using the shared VAT category)
+		{
+			Code: tax.CategoryVAT,
```
samlown: Repeating above, I suspect Japan needs its own type of tax: `JCT`.

regimes/jp/tax_categories.go
```diff
@@ -0,0 +1,95 @@
...
+			Rates: []*tax.RateDef{
+				{
+					Keys: []cbc.Key{tax.KeyStandard},
+					Rate: tax.RateGeneral,
+					Name: i18n.String{
+						i18n.EN: "Standard rate",
+						i18n.JA: "標準税率",
+					},
+					Description: i18n.String{
+						i18n.EN: "The standard rate of Consumption Tax applies to all supplies of goods and services unless a reduced rate, zero rate, or exemption applies. The 10% headline rate combines a 7.8% national Consumption Tax and a 2.2% Local Consumption Tax.",
```
samlown: Again, I'd probably avoid adding the percentage values themselves.

regimes/jp/tax_categories.go
```diff
@@ -0,0 +1,95 @@
...
+				{
+					Keys: []cbc.Key{tax.KeyStandard},
+					Rate: tax.RateReduced,
+					Name: i18n.String{
+						i18n.EN: "Reduced rate",
+						i18n.JA: "軽減税率",
+					},
+					Description: i18n.String{
+						i18n.EN: here.Doc(`
+							The 8% reduced rate applies to food and drink for human
```
samlown: Same here, no 8%, just "The reduced rate applies..."

samlown (changes requested):
Good from outset. A few more examples would help, and descriptions that don't include data. I think the tax category needs to be something that reflects the Japanese approach, which is not quite the same as regular VAT nor GST.

regimes/jp/jp.go
fenomenog4: Makes sense 👍

regimes/jp/jp.go
fenomenog4: Yeah, it's JCT, but as far as I can tell, "Consumption Tax" is structurally the same as VAT, so I reused the shared `tax.CategoryVAT` and just localised the name. Same call ES made for IVA. The alternative would be a bespoke JCT category, but reusing the shared one keeps it in the common machinery, and it still has to serialise to VAT in later export steps anyway.

As for the TaxScheme field itself: as I've traced it, that's the wire code that ends up in `cac:TaxScheme/cbc:ID` when the invoice serialises to UBL, and JP PINT is strict that it must be VAT. The official document include assertions for exactly that, and their reference invoices use VAT everywhere. Context: https://docs.peppol.eu/poac/jp/pint-jp/resources.zip and https://docs.peppol.eu/poac/jp/pint-jp/trn-invoice/rule/aligned-ibr-jp-03/ 

That said, if I'm wrong, or if you'd rather have a dedicated JCT category code, or think the JCT identity belongs in an extension, I'm happy to take another pass. Just let me know which way you're leaning.

regimes/jp/tax_categories.go
fenomenog4: Added an answer to the other comment

regimes/jp/jp.go
fenomenog4: Update: I'm still mulling this over. I'll try to come up with something and let you know when it's ready.

fenomenog4:
Hi, @samlown. Thanks for the review. Here's an update on the comments:

- I have added more examples, which now cover foreign currency, discounts/charges, reverse charges, simplified taxation, credit notes, exports, and payment instructions. There are a total of 8 examples now.
- The descriptions for the regime and rates no longer hard-code specific figures.
- I fixed the grammar in the tax identity comment (Copilot).
- Consumption Tax now has its own dedicated JCT category instead of reusing `tax.CategoryVAT`. However, `RegimeDef.TaxScheme` still remain as VAT since JP PINT strictly requires `cac:TaxScheme/cbc:ID` to use that value. This is the same split regimes/in uses; `RegimeDef.TaxScheme` is set to GST but then custom categories are defined. If you still have reservations here or would rather it went a different way, let me know and I'll make another pass.
- The current golangci-lint failures are unrelated to this pull request and will resolve once I merge with the main branch and address any conflicts.

PR #949, opened 2026-09-01
Allow attributes identified only by a label

mrdanwa:
- `org`: `Attribute` no longer requires a `key` or `type` when a `label` is present. UBL imports (e.g. France) map each `AdditionalItemProperty` to an attribute with only `label` and `text`, which failed `GOBL-ORG-ATTRIBUTE-01` since the label is presentation-only.
- `GOBL-ORG-ATTRIBUTE-01` now only rejects attributes that set both `key` and `type`; the new `GOBL-ORG-ATTRIBUTE-04` ("attribute must have a key, a type, or a label") is only checked when there is no `label`, so attributes such as `{ "label": "SubitemValue3", "text": "850,36" }` are now valid.
- The identification rules and the large value rule (`02`) use named `is.Func` predicates, following the prevailing style in the codebase; the small unit rule (`03`) expressions are untouched.
- Updated struct docs, regenerated `data/rules/org.json` and `data/schemas/org/attribute.json`, and added a changelog entry.

## Pre-Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Changed
 
+- `org`: `Attribute` no longer requires a `key` or `type` when a `label` is present.
 - `net`: signatures are no longer interpreted by position. `Client.VerifyEnvelope`
   is replaced by `Client.VerifyParty`, which verifies the address declared by a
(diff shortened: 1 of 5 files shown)


samlown (approved):
Makes sense. As discussed, I'd swap the big Expr in org/attribute for a function call for coherence.

PR #955, opened 2026-09-03
bill: use a value alias in Invoice.UnmarshalJSON

### LINKED ISSUE #954: bill.Invoice.UnmarshalJSON: infinitely recurses under Go 1.27 (jsonv2)
## Summary

`(*bill.Invoice).UnmarshalJSON` uses `type Alias *Invoice` to strip its own method before delegating to `json.Unmarshal`. That trick works under the `encoding/json` v1 implementation but not under the v2 implementation that Go 1.27 enables by default, where the method is found again and the function calls itself until the stack overflows.

Any `json.Unmarshal` into a `*bill.Invoice` crashes the process on Go 1.27. This is not recoverable by the caller. `fatal error: stack overflow` is not a panic and cannot be caught with `recover()`.

## Version

- gobl: `main` as of 2885bc98e2c7c5e42706550ac5b97ddf87c20ffa
- Go: `go1.27.0 darwin/arm64`

## Reproduction

gobl's own test suite reproduces it — no external code needed:

```
git clone https://github.com/invopop/gobl && cd gobl
go test ./bill/...
```

Result on Go 1.27:

```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

goroutine 302 gp=0x437221f92000 m=0 mp=0x103702140 [running]:
encoding/json/internal/jsonopts.(*Struct).Join(...)
encoding/json/jsontext.(*decoderState).reset(...)
encoding/json/v2.Unmarshal(...)
encoding/json.Unmarshal(...)
	/usr/local/go/src/encoding/json/v2_decode.go:111
...
github.com/invopop/gobl/bill_test.TestInvoiceUnmarshalJSON
	bill/invoice_test.go:1163
FAIL	github.com/invopop/gobl/bill	1.309s
```

The same command passes with the experiment disabled:

```
GOEXPERIMENT=nojsonv2 go test ./bill/...
```

## Root cause

bill/invoice.go:

```go
func (inv *Invoice) UnmarshalJSON(data []byte) error {
	type Alias *Invoice
	if err := json.Unmarshal(data, (Alias)(inv)); err != nil {
```

`Alias` is a defined *pointer* type. Under v1, a defined pointer type has an empty method set, so `json.Unmarshal` does not see `UnmarshalJSON` and decodes the struct fields directly — the intended behaviour.

Under v2, the decoder dereferences through the named pointer type before checking for the `Unmarshaler` interface, finds `(*Invoice).UnmarshalJSON` again, and calls it. Each call re-enters itself with the same data.

Go 1.27 turns the `jsonv2` experiment on by default, so plain `encoding/json` is now the v1-on-v2 implementation and the pattern breaks.

This behaviour difference is tracked upstream in golang/go as golang/go#75361. The Go side may or may not change, but the pattern is fragile either way and the fix on gobl's side is independent of it.

## Proposed fix

Use the conventional value-typed alias, which has an empty method set under both implementations:

```go
	type Alias Invoice
	if err := json.Unmarshal(data, (*Alias)(inv)); err != nil {
```

This is the form already used everywhere else in the codebase (`bill/tax.go`, `bill/payment.go`, `pay/terms.go`, `pay/instructions.go`,
`pay/record.go`, `tax/combo.go`). `bill/invoice.go` is the only pointer-alias site.

## Workaround

Downstream projects on Go 1.27 can build and test with `GOEXPERIMENT=nojsonv2` to avoid the crash. `nojsonv2` will go away, tho.

d--j:
fixes #954 

## Problem

`(*Invoice).UnmarshalJSON` strips its own method with `type Alias *Invoice` before delegating to `json.Unmarshal`. A defined *pointer* type has an empty method set under the `encoding/json` v1 implementation, so this works there.

Under the v2 implementation — enabled by default in Go 1.27 — the decoder dereferences through the named pointer type before checking for `json.Unmarshaler`, finds `(*Invoice).UnmarshalJSON` again, and re-enters it with the same data. Every `json.Unmarshal` into a `*bill.Invoice` recurses until `fatal error: stack overflow`, which callers cannot recover from.

gobl's own `TestInvoiceUnmarshalJSON` crashes on Go 1.27 today.

## Change

One line, in `bill/invoice.go`:

```diff
 func (inv *Invoice) UnmarshalJSON(data []byte) error {
-	type Alias *Invoice
-	if err := json.Unmarshal(data, (Alias)(inv)); err != nil {
+	type Alias Invoice
+	if err := json.Unmarshal(data, (*Alias)(inv)); err != nil {
 		return err
 	}
```

A value-typed alias has an empty method set under both implementations. This is the form already used at every other alias site in the codebase (`bill/tax.go`, `bill/payment.go`, `pay/terms.go`, `pay/instructions.go`, `pay/record.go`, `tax/combo.go`); `bill/invoice.go` was the only outlier.

## A note on embedded types

`type Alias *Invoice` has a genuinely empty method set. `type Alias Invoice` does not: methods promoted from embedded fields are part of the new type's method set.

`Invoice` embeds `tax.Regime`, `tax.Addons`, `tax.Tags` and `uuid.Identify`. None of them declares `UnmarshalJSON`, so nothing is promoted and the recursion does not reappear by another route. It is a constraint a future embed could silently break, so I am happy to add a comment on the `type Alias` line — or a guard test — if you would like one.

## Testing

On `go1.27.0` (jsonv2 on by default):

- `go test ./...` — passes. Fails with `fatal error: stack overflow` before this change.

With the experiment disabled:

- `GOEXPERIMENT=nojsonv2 go test ./...` — passes. No regression against the v1 implementation.

Also verified against a downstream consumer (a UBL/CII e-invoicing service): its suite crashes with unpatched gobl under Go 1.27 and passes with this change, with no `GOEXPERIMENT=nojsonv2` workaround needed.

## Why it matters

Go 1.27 makes this the default path. Downstream projects currently have to build and test with `GOEXPERIMENT=nojsonv2` to use gobl at all, and that flag will not exist forever.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,8 +8,24 @@ on:
   pull_request:
 jobs:
   test:
-    name: Test
+    name: Test (Go ${{ matrix.go-version }})
     runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # The three most recent Go releases. The oldest matches the go
+        # directive in go.mod, and 1.27 is the first release to enable the
+        # jsonv2 implementation of encoding/json by default.
+        go-version: ["1.25.x", "1.26.x", "1.27.x"]
+        include:
+          - go-version: "1.27.x"
+            coverage: true
+
+    env:
+      # Run on the toolchain the matrix installed instead of downloading a
+      # newer one when a module in the graph asks for it.
+      GOTOOLCHAIN: local
 
     steps:
       - name: Check out code
@@ -18,7 +34,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version-file: "go.mod"
+          go-version: ${{ matrix.go-version }}
 
       - name: Install Dependencies
         env:
@@ -33,6 +49,7 @@ jobs:
         run: go test -race -coverprofile=coverage.out -covermode=atomic -coverpkg=./... ./...
 
       - name: Upload coverage reports to Codecov
+        if: matrix.coverage
         uses: codecov/codecov-action@v5.1.2
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
   signature entries instead of panicking.
 - `regimes/ar`: CUIT/CUIL tax identities with the `24` prefix (an individual
   contingency prefix) are no longer wrongly rejected as having an invalid prefix.
+- `bill`: `Invoice.UnmarshalJSON` no longer recurses infinitely under the `jsonv2` implementation of `encoding/json`, enabled by default from Go 1.27. The method used a defined pointer type (`type Alias *Invoice`) to strip itself before delegating, which has an empty method set under the v1 implementation but not under v2, where the decoder dereferences through the named pointer and finds `UnmarshalJSON` again. Unmarshalling any invoice on Go 1.27 ended in `fatal error: stack overflow`, which callers cannot recover from. The alias is now value-typed (`type Alias Invoice`), matching every other alias site in the codebase and behaving identically under both implementations.
 
 ## [v0.504.0]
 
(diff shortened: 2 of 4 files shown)


samlown (changes requested):
Many thanks for identifying this and suggesting a fix! There may be other places in the GOBL code base where this happens and I think we should also have some tests in place to ensure it doesn't happen in the future. I'll take a look.

samlown (approved):
Added some tests, looks great. Merging. Thanks!

PR #957, opened 2026-09-07
cal: accept a single bound on Period (EN 16931 BR-CO-19 / BR-CO-20)

methodofaction:
Resolves [GBL-93](https://linear.app/invopop/issue/GBL-93/calperiod-accept-a-single-bound-to-match-en-16931-br-co-19-br-co-20).

## Problem

`cal.Period` required both a start and an end date, but EN 16931 only requires at least one of BT-73/BT-74 (BR-CO-19) and BT-134/BT-135 (BR-CO-20). Conformant inbound invoices carrying a one-sided period (e.g. the STEF Transport case from Pylon #6193) were converted into GOBL documents that could not be validated or signed.

## Changes

- `cal.Period`: **breaking**: `Start` and `End` are now `*Date` pointers, so an absent bound is `nil` and omitted from JSON instead of being serialized as `"0000-00-00"`. Legacy `"0000-00-00"` input still parses.
- At least one of `start` / `end` is required. A period with neither fails once, on the period itself, with the new `GOBL-CAL-PERIOD-11`.
- `GOBL-CAL-PERIOD-01` / `-02` ("date cannot be zero") are kept, but as `rules.AssertIfPresent`: they only apply to a bound that is actually supplied, so an explicit zero date is still rejected while a missing one is not.
- Rule `10` (end on or after start) applies only when both bounds are present and non-zero, so a zero date reports one fault rather than two.
- Regenerated `data/schemas/cal/period.json` (no longer requires both fields) and `data/rules/cal.json`.
- New FR example fixture `examples/fr/invoice-fr-fr-period.yaml` with an end-only header period and start-only line periods, mirroring the reported invoice.
- CHANGELOG entry.

## Verified against the reported invoice

Parsing the actual STEF UBL file with `gobl.ubl`: on v0.504.0 the only faults are the period codes (header start, all 11 line ends). On this branch the document validates cleanly under the FR regime with the EN 16931 and FR CTC flow 2 addons. That check was run before the switch to pointer dates; the accepted-document behavior is unchanged, only the fault code for a bound-less period moved from `-03` to `-11`.

## Consumer audit

No code in `bill`, `org`, `tax`, `pay`, `addons` or `regimes` reads `Period.Start` / `Period.End` directly; the only reference is a nil check in `SubLine.IsEmpty`. Regimes whose authorities require both bounds (e.g. Argentina ARCA for service invoices) previously relied on this core rule implicitly and may want an explicit "both required" rule; tracked as a follow-up.

## Companion PRs

- invopop/gobl.cii#94: guards on the CII exporter plus round-trip fixtures.
- invopop/gobl.ubl#140: round-trip fixtures (exporter already safe).

Both companions were written against the value-typed fields and need updating for the pointer change before they merge.

### DIFF
--- a/cal/period.go
+++ b/cal/period.go
@@ -21,26 +21,26 @@ type Period struct {
 
 func periodRules() *rules.Set {
 	return rules.For(new(Period),
-		rules.When(is.Expr(`End.IsZero()`),
-			rules.Field("start",
-				rules.Assert("01", "start date cannot be zero",
-					DateNotZero(),
-				),
-			),
-		),
-		rules.When(is.Expr(`Start.IsZero()`),
-			rules.Field("end",
-				rules.Assert("02", "end date cannot be zero",
-					DateNotZero(),
-				),
-			),
-		),
 		rules.Object(
+			rules.Assert("03", "either a start or end date is required",
+				is.Func("has start or end", periodHasBound),
+			),
 			rules.Assert("10", "end date must be on or after start date",
 				is.Func("end not before start", periodEndNotBeforeStart),
 			),
 		),
 	)
+}
+
+// periodHasBound checks that at least one of the two dates is set. Which one
+// is irrelevant: EN 16931 BR-CO-19 and BR-CO-20 allow either bound alone, so
+// this is reported once against the period rather than against each field.
+func periodHasBound(val any) bool {
+	p, ok := val.(*Period)
+	if !ok || p == nil {
+		return true
+	}
+	return !p.Start.IsZero() || !p.End.IsZero()
 }
 
 func periodEndNotBeforeStart(val any) bool {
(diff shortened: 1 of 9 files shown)


cal/period.go
```diff
@@ -5,26 +5,34 @@ import (
 	"github.com/invopop/gobl/rules/is"
 )
 
-// Period represents two dates with a start and finish.
+// Period represents a span of time bounded by a start and/or an end date.
+// At least one of the two bounds must be provided, but either may be omitted
+// on its own, mirroring EN 16931 (BR-CO-19 and BR-CO-20) where the invoicing
+// and line period start and end dates are each optional.
```
samlown: A bit too much implementation specific detail here.

samlown (approved):
LGTM

PR #964, opened 2026-09-14
bill: add RoundToCurrency and respect explicit rounding rules

samlown:
## Context

EN 16931's `BR-DEC-*` rules cap nearly every amount at the currency's precision, while `BR-CO-10` requires the lines to sum to the total. GOBL's default `precise` rounding rule satisfies neither: line totals can carry extra decimals, and rounding each field on the way out stops them adding up.

Since v0.505 `RemoveIncludedTaxes` also switches a document from `currency` to `precise`, which made this reachable for any tax-inclusive document.

As a result **gobl.ubl and gobl.cii each wrote the same ~110-line fix, byte-for-byte identically** (`gobl.ubl/ubl.go`, `gobl.cii/rounding.go`), and two addons (`gobl.dk.oioubl`, `gobl.fr.ctc`) solve it a third way by forcing `currency` during normalization — only for `RemoveIncludedTaxes` to silently revert them.

This moves the mechanism into GOBL so the two approaches become first-class and composable:

```go
// exact taxes, precise amounts
inv.RemoveIncludedTaxes()

// every amount fits the currency, drift in Totals.Rounding (BT-114)
inv.RemoveIncludedTaxes()
inv.RoundToCurrency()

// also valid alone: a precise invoice with no included taxes
// still breaks BR-DEC-23 when 3 x 10.555 = 31.665
inv.RoundToCurrency()
```

The two are a genuine trade-off, not redundant: `precise` preserves the tax actually charged (`84.91`), `currency` guarantees every amount is expressible (`84.90` + `0.01` rounding). The format decides.

## Changes

- **Added** `Invoice.RoundToCurrency()` (`bill/rounding.go`). No-op when the document already fits, so nothing is recalculated unnecessarily.
- **Changed** `RemoveIncludedTaxes` to only switch to `precise` when the rule was inherited from the regime (Greece). A rule set on the document is respected.
- **Fixed**, via the now-shared residual handling: the residual is anchored on `Payable` rather than `TotalWithTax` (retained-tax drift previously escaped into `Due` and percentage advances); an existing `Totals.Rounding` is added to rather than replaced (a user-supplied BT-144 was being discarded); and no more panic on an uncalculated document carrying `prices_include`.

Also guards against `tax.TagBypass`, and rescales `Discount.Base`/`Charge.Base` — which the downstream copies were missing. Those are never reduced by the calculator and `round()` keys its precision off them, so a 4dp base made the precision check true forever while recalculation could never resolve it.

No new public struct fields.

### Unrelated drive-by

The last two commits mark the delivery `type` as calculated in the JSON schema. The hardcoded enum extra was being rendered as `"enum": "advice"`; the type list already comes from the `JSONSchemaExtend` one-of, matching invoices.

## Verification

Full suite, `go vet` and `golangci-lint` clean, with **no `examples/**/out/*.json` drift**.

The meaningful check: in scratch copies of gobl.ubl and gobl.cii pointed at this branch, I deleted their local `roundToCurrency`/`exceedsCurrencyPrecision` and swapped in `doc.RoundToCurrency()`. Both suites pass with **zero golden XML changes**. That confirms the pairing reproduces the pre-v0.505 numbers exactly (`1415.04 / 84.90 / 0.06 / 1500.00`), which is locked in as a regression test.

## Follow-ups (deliberately not bundled)

1. `RemoveIncludedTaxes` doesn't strip `Discount.Base`, `Charge.Base` or `LineCharge.Rate`, so a percentage discount with an explicit base stays tax-inclusive after removal.
2. `Invert()` and `ConvertInto()` both nil `Totals`, discarding the BT-114 residual; `Invert` then fails its own check on any invoice carrying one. `ConvertInto` also shallow-copies the `Tax` pointer.
3. The precision check is a decimal-count heuristic. A document can be all-2dp and still break BR-CO-10 (two lines of qty 0.5 x 10.05 display 5.03 + 5.03 = 10.06 while `Sum` is 10.05).

Downstream migrations for gobl.ubl / gobl.cii / the addons want their own PRs after a release.


samlown:
Merged `main` and moved this PR onto the new change file layout from #966.

All four entries moved from `CHANGELOG.md` to `changes/unreleased/bill-round-to-currency.md`, text unchanged — the `RoundToCurrency` addition, the `RemoveIncludedTaxes` rule change, and both fixes. `CHANGELOG.md` is taken from `main` verbatim, since it is now generated.

`go generate` produced no drift, lint passes, and the full suite passes locally.

cavalle (approved):
LGTM

PR #965, opened 2026-09-14
bill: gather the rounding policy into bill/rounding.go

samlown:
Stacked on #964 — review that one first. This commit is a pure move, no behaviour change.

## Why

Three precision constants were declared in files that never used them:

| constant | was declared in | actually used by |
|---|---|---|
| `defaultTaxRemovalAccuracy` | `invoice.go` | `charges.go`, `discounts.go`, `line.go` |
| `defaultCurrencyConversionAccuracy` | `invoice.go` | `exchange.go` only |
| `linePrecisionExtra` | `line_calculate.go` | also `charges.go`, `discounts.go` |

The first and third are both `2` but mean different things — one is the `precise` rule's extra digits, the other the headroom for dividing tax back out of a price. They now sit together with a comment saying which is which, so the next person to touch one doesn't assume they're the same knob. (There is a third instance of the same idea at `tax/totals_calculator_precise.go:13`, left alone as it crosses a package boundary.)

`roundingRule` and `recalculateKeepingPayable` were added to `calculator.go` in #964 only because that is where the inlined copies were. Both are rounding policy rather than calculation, so they move too.

`defaultCurrencyConversionAccuracy` goes to `exchange.go` instead — it is not a rounding concern.

## What deliberately did not move

The per-type `round` methods: `Charge.round`, `Discount.round`, `Line.round`, `SubLine.round`, `Totals.round`, and the `roundLines`/`roundDiscounts`/`roundCharges` helpers.

Two reasons. The package organises by type — `charges.go` keeps Charge's validation, calculation, rounding and tax removal together — and pulling only the rounding out would break that for every type at once. And they are a different concern from this file: all four are the final pass *inside* `calculate`, via `roundTotalsAndPreparePayments`, whereas `rounding.go` holds a document-level operation the caller invokes, closer to `invoice_convert.go` than to the calculator.

## Verification

- `go doc -all ./bill` is byte-identical before and after, so nothing about the public API moved.
- Full suite, `go vet` and `golangci-lint` clean; `bill` coverage unchanged at 93.8%, which is what a pure move should do.


samlown:
Merged the updated #964 branch, which brings in the new change file layout from #966. The merge was clean — no conflicts.

No change file added here on purpose: this PR only moves the rounding policy between files in `bill/`, with no change visible to anyone using GOBL, so a release note entry would be noise. The entries for the behaviour this builds on are in #964 at `changes/unreleased/bill-round-to-currency.md`.

`go generate` produced no drift, lint passes, and the full suite passes locally.

PR #975, opened 2026-09-15
c14n: accept U+FFFD as valid content

alvarolivie:
`utf8.DecodeRuneInString` reports `RuneError` in two different cases, told apart only by size:

| input | size | meaning |
|---|---|---|
| malformed bytes, e.g. `0xE9` | 1 | genuinely invalid UTF-8 |
| U+FFFD, encoded `EF BF BD` | 3 | a valid code point |

`encodeString` checked only the rune, so any document containing a replacement character failed to digest with a misleading `json: unsupported value` — even though the string is perfectly valid UTF-8 and `encoding/json` marshals it without complaint.

This turned up on a French CTC invoice whose sender had already mangled its own text (a Latin-1 → UTF-8 conversion gone wrong, leaving `Premi<U+FFFD>re v<U+FFFD>rification`). The mojibake is the sender's, but it made the whole document unparseable rather than merely ugly.

Adding the usual `&& size == 1` companion check restores the intended meaning.

### What this actually changes

`CanonicalJSON` unmarshals before it marshals, and `encoding/json` replaces malformed bytes with U+FFFD during that decode. By the time `encodeString` runs no raw malformed byte has survived — every path through `CanonicalJSON`, including `Envelope.Digest`, sees valid UTF-8 already.

So the honest summary is: **gobl no longer rejects documents whose text was laundered to U+FFFD.** That covers both text a sender mangled upstream and text `encoding/json` normalized on the way in. At the point the check runs the two are indistinguishable — and were before this PR too.

The corollary is that gobl no longer trips on malformed UTF-8 at the `c14n` layer. It never meaningfully detected it there: `encoding/json` had already normalized the bytes, irreversibly, before the check ran, so the rejection was incidental rather than a real integrity guard. A genuine tripwire belongs on the raw input, ahead of JSON decoding, not here.

The `&& size == 1` branch stays reachable only when a `c14n.String` is constructed directly from a Go string holding invalid bytes — which is what the existing `\xff\xfe` test does, and it still passes.

Digest/content consistency is unaffected: the digest is computed over the same marshalled bytes that get stored.

Verified end-to-end: a converter built against this branch parses the invoice above, preserving the U+FFFD as content rather than substituting anything. Full suite green.


alvarolivie:
Closing in favour of handling this in the converters. Cleaning at the field level in gobl.cii / gobl.ubl covers the decoded text directly, without loosening a check in gobl's canonical JSON.

Worth recording the finding regardless: the `utf8.RuneError` check in `c14n/models.go` cannot distinguish malformed bytes (size 1) from a correctly encoded U+FFFD (size 3), so it rejects a valid code point. It also runs after `encoding/json` has already rewritten malformed bytes to U+FFFD, so it never detected corruption in the first place.

